PR #156, opened 2019-08-16
Feature/vih 4677 user sees message if they use an unsupported browser

RocheL:
VIH-4677
### Change description ###
pending IE 11 : added the check for IE and html for the unsupported browser page in index.html
```
[ ] Yes
[x] No
```


VideoWeb/VideoWeb/ClientApp/src/index.html
```diff
@@ -19,13 +19,145 @@
 
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <link rel="icon" type="image/x-icon" href="favicon.ico">
+  <script>
```
shaed-parkar: This should navigate to the not supported page rathe than duplicate the whole page

PR #398, opened 2020-02-27
Feature/vih 5427 reconnect to eventhub

shaed-parkar:

[VIH-5427](https://tools.hmcts.net/jira/browse/VIH-5427)

### Change description ###

* check if video needs to be shown/hidden after eventhub reconnect

```
[ ] Yes
[x] No
```


VideoWeb/VideoWeb/ClientApp/src/app/services/api/video-web.service.ts
```diff
@@ -36,6 +36,7 @@ export class VideoWebService {
     }
 
     getConferenceById(conferenceId: string): Observable<ConferenceResponse> {
+        console.trace('boop de bop');
```
RuwanikaWeerakkody: What is this Shaed :)

VideoWeb/VideoWeb/ClientApp/src/app/shared/beta-banner/beta-banner.component.ts
```diff
@@ -1,56 +1,61 @@
...
+    }
+    ngOnInit(): void {
+        this.getInPageFeedbackUrl();
+        this.setupSubscribers();
+    }
 
-  updateFeedbackUrl(feedbackUrl: string): void {
-    this.pageUrl = feedbackUrl + this.router.url;
-  }
+    ngOnDestroy(): void {
```
RuwanikaWeerakkody: Why are we doing this in beta banner?

VideoWeb/VideoWeb/ClientApp/src/app/services/api/video-web.service.ts
shaed-parkar: haha i needed to figure out which test was calling the API rather thana stub

shaed-parkar: removed :)

VideoWeb/VideoWeb/ClientApp/src/app/shared/beta-banner/beta-banner.component.ts
shaed-parkar: we need to manage our subscriptions. I noticed this when i was digging through tests calling the API. If we don't unsubscribe between components then we will get memory leaks. We should either convert these api calls to promises or unsubscribe when the component is destroyed

PR #431, opened 2020-03-13
Stop Judge disconnect events being published to clients

shaed-parkar:

[VIH-5677](https://tools.hmcts.net/jira/browse/VIH-5677)

### Change description ###

There is no need to publish disconnect events for judges since the Judge Available, Judge Unavailable events manage the concern about a judge's availability. This should stop the latter events from being overwritten due to the time is takes for Kinly to notify a judge has been disconnected.

```
[ ] Yes
[x] No
```

### DIFF
--- a/VideoWeb/VideoWeb.EventHub/Handlers/DisconnectedEventHandler.cs
+++ b/VideoWeb/VideoWeb.EventHub/Handlers/DisconnectedEventHandler.cs
@@ -19,7 +19,8 @@ namespace VideoWeb.EventHub.Handlers
 
         protected override async Task PublishStatusAsync(CallbackEvent callbackEvent)
         {
-            await PublishParticipantDisconnectMessage().ConfigureAwait(false);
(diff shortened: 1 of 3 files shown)


VideoWeb/VideoWeb.EventHub/Handlers/DisconnectedEventHandler.cs
```diff
@@ -19,7 +19,8 @@ public DisconnectedEventHandler(IHubContext<Hub.EventHub, IEventHubClient> hubCo
 
         protected override async Task PublishStatusAsync(CallbackEvent callbackEvent)
         {
-            await PublishParticipantDisconnectMessage().ConfigureAwait(false);
+            // skip disonnect events since judge available/unavailable statuses are tracked by other events
```
RuwanikaWeerakkody: Hi Shaed, Can we look at why the original issue is happening first before disabling this?

PR #467, opened 2020-03-27
VIH-5731 update to use new contracts from video api

shaed-parkar:

VIH-5731

### Change description ###

use new judge and indiviual endpoints

```
[x] Yes
[ ] No
```


VideoWeb/VideoWeb.Contract/Responses/ConferenceStatus.cs
```diff
@@ -1,14 +1,14 @@
...
-    {
-        NotStarted = 0,
-        InSession = 1,
-        Paused = 2,
-        Suspended = 3,
-        Closed = 4
-    }
-}
\ No newline at end of file
+// namespace VideoWeb.Contract.Responses
```
rehmanab: Commented out code

VideoWeb/VideoWeb/Controllers/ParticipantsController.cs
```diff
@@ -85,6 +90,20 @@ public async Task<IActionResult> GetTestCallResultForParticipantAsync(Guid confe
...
+        {
+            var conference = _conferenceCache.GetConference(conferenceId);
+            if (conference == null)
+            {
+                var conferenceDetail = await _videoApiClient.GetConferenceDetailsByIdAsync(conferenceId);
+                await _conferenceCache.AddConferenceToCache(conferenceDetail);
+                conference = _conferenceCache.GetConference(conferenceId);
+            }
+
+            return conference.Participants
```
rehmanab: Might throw here... Just in case there are no Participants, which will probably never going to happen, but could be some defensive coding ??

shaed-parkar: If it doesn't exist, it will throw an exception because of the Single() and not SingleOrDefault() method

VideoWeb/VideoWeb.Contract/Responses/ConferenceStatus.cs
shaed-parkar: cheers!

PR #509, opened 2020-04-13
VIH-5762 Fix HubConnection cannot start

valentinabennett:

https://tools.hmcts.net/jira/browse/VIH-5781

### Change description ###
Fix for case HubConnection cannot start in not Disconnected state

```
[ X] Yes
[ ] No
```


valentinabennett:
We set automatic reconnect options in the configuration for HubConnection(repeat 7 times in period of msec), and the re connection started automatically without this code. (I have tested it). And if re-connection started we manually start reconnect as well, that gives the error that HubConnection is not in Disconnected state. So this code duplicate auto re connection.

RuwanikaWeerakkody:
> We set automatic reconnect options in the configuration for HubConnection(repeat 7 times in period of msec), and the re connection started automatically without this code. (I have tested it). And if re-connection started we manually start reconnect as well, that gives the error that HubConnection is not in Disconnected state. So this code duplicate auto re connection.

I agree Valentina.

PR #514, opened 2020-04-15
VIH-5825 Cache users

shaed-parkar:

VIH-5825

### Change description ###

cache token and add user profile to claims

```
[ ] Yes
[x] No
```

### DIFF
--- a/VideoWeb/VideoWeb.Common/Extensions/EnumExtensions.cs
+++ b/VideoWeb/VideoWeb.Common/Extensions/EnumExtensions.cs
@@ -2,7 +2,7 @@ using System.ComponentModel;
 using System.Reflection;
 using System.Runtime.Serialization;
 
-namespace VideoWeb.Extensions
+namespace VideoWeb.Common.Extensions
 {
     public static class EnumExtensions
     {
--- a/VideoWeb/VideoWeb.Common/Extensions/EnumExtensions.cs
+++ b/VideoWeb/VideoWeb.Common/Extensions/EnumExtensions.cs
@@ -1,5 +1,4 @@
 using System.ComponentModel;
-using System.Reflection;
 using System.Runtime.Serialization;
 
 namespace VideoWeb.Common.Extensions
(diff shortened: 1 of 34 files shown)


VideoWeb/VideoWeb.Common/SignalR/AdUserProfileService.cs
```diff
@@ -5,8 +5,8 @@ namespace VideoWeb.Common.SignalR
 {
     public interface IUserProfileService
     {
-        Task<bool> IsVhOfficerAsync(string username);
         Task<string> GetObfuscatedUsernameAsync(string username);
+        // Task<UserProfile> GetUserAsync(string userName);
```
rehmanab, line 9: comment

VideoWeb/VideoWeb/Controllers/ProfilesController.cs
```diff
@@ -29,20 +36,26 @@ public ProfilesController(IUserApiClient userApiClient, ILogger<ProfilesControll
...
         {
-            var username = User.Identity.Name.ToLower().Trim();
             try
             {
-                var profile = await _userApiClient.GetUserByAdUserNameAsync(username);
-                var response = UserProfileResponseMapper.MapToResponseModel(profile);
+                var role = User.Claims.First(c => c.Type == ClaimTypes.Role).Value;
+                var response = new UserProfileResponse
+                {
+                    FirstName = User.Claims.First(c => c.Type == ClaimTypes.GivenName).Value,
```
rehmanab, line 47: Can throw exceptions

VideoWeb/VideoWeb.Common/Extensions/EnumExtensions.cs
```diff
@@ -2,7 +2,7 @@
 using System.Reflection;
 using System.Runtime.Serialization;
 
-namespace VideoWeb.Extensions
+namespace VideoWeb.Common.Extensions
 {
     public static class EnumExtensions
     {
```
rehmanab, line 8: Remove the else as its not needed as your returning result above

rehmanab: Remove attributes != null as its always true

rehmanab: Change to vars instead of explicitly typed

VideoWeb/VideoWeb/Controllers/ProfilesController.cs
shaed-parkar: If we do this then we need to make a few more  changes to the the web app to assume a 500 response code means user is not authorised. Think that might be beyond the scope of this but I don't  mind doing it. Just worried about the domino effect

PR #519, opened 2020-04-17
VIH-5846 Add filter for vho conferences

shaed-parkar:

vih-5846

### Change description ###

* Add venue filter for VHOs 

* Also took out the use  of TestBeds in  the vho components and security components since we only have a handful of tests that actually use the DOM
  * This makes it much easier to test functions/components/services individually rather than having the whole angular life cycle execute and having to code around it
  * Tests significantly faster (the goal is to remove testbeds completely so that mutation tests do not take 13+ hours to run)
  * Exposes functions that do  not have tests for it - as a result i’ve added ~ 70 tests around VHO components

```
[ ] Yes
[x] No
```


shaed-parkar:
azp /run

PR #633, opened 2020-06-08
VIH-5405 set for participant not signed in status

valentinabennett:

https://tools.hmcts.net/jira/browse/VIH-5405

### Change description ###
Set NotSignedIn status for participant on signed off

```
[ X] Yes
[ ] No
```


VideoWeb/VideoWeb/ClientApp/src/app/services/participant-status-update.service.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+import { Logger } from 'src/app/services/logging/logger-base';
+import { TestBed, fakeAsync } from '@angular/core/testing';
+
+class MockRouter {
+    public url = '/check-equipment/1234-1234-1234';
+}
+
+describe('ParticipantStatusUpdateService', () => {
+
+    let videoWebServiceSpy: jasmine.SpyObj<VideoWebService>;
```
shaed-parkar, line 14: Could you remove the test bed here and follow suite like the other tests. Thanks

VideoWeb/VideoWeb/ClientApp/src/app/services/participant-status-update.service.ts
```diff
@@ -0,0 +1,46 @@
...
+    pageUrls.MicrophoneWorking,
+    pageUrls.VideoWorking,
+    pageUrls.HearingRules,
+    pageUrls.Declaration
+];
+
+@Injectable({
+    providedIn: 'root'
+})
+export class ParticipantStatusUpdateService {
```
shaed-parkar, line 23: Wrapping a service to call another service adds a layer of complexity we do not need.

1. Either remove this service and call videoWebService 
or 
2. Call the apiClient directly in this service

VideoWeb/VideoWeb/ClientApp/src/app/services/participant-status-update.service.spec.ts
valentinabennett: done

VideoWeb/VideoWeb/ClientApp/src/app/services/participant-status-update.service.ts
valentinabennett: done

VideoWeb/VideoWeb/ClientApp/src/app/services/participant-status-update.service.spec.ts
valentinabennett: done

VideoWeb/VideoWeb/ClientApp/src/app/app.component.ts
```diff
@@ -145,4 +150,23 @@ export class AppComponent implements OnInit, OnDestroy {
...
+
+    @HostListener('window:beforeunload', ['$event'])
+    async beforeunloadHandler($event: any) {
+        $event.preventDefault();
+        await this.participantStatusUpdateService.postParticipantStatus(EventType.ParticipantNotSignedIn);
+        $event.returnValue = 'save';
+    }
+
+    refreshPageParticipant() {
+        this.subscriptions.add(this.router.events.subscribe(async (event) => {
```
shaed-parkar, line 162: It looks like the app component is subscribing to the same observable twice. Once here and once in `scrollToTop`.

There should be a function per each bit of logic that needs to execute on a `NavigationEnd` event. This will make it easier to track what behaviour is executing against this observable

valentinabennett: If I subscribe to catch refresh from router event  in ngOnInit , it does not work. In constructor it does work. But scrolling should be in ngOnInit as it depends on DOM element.

PR #686, opened 2020-06-23
VIH-5255 Remove test for Edge from overnights

npoxon:
### Change description ###
- Edge is supported but not by automation tests. Removing from overnight tests

```
[ ] Yes
[X] No
```

### DIFF
--- a/azure-pipelines.nightly.yml
+++ b/azure-pipelines.nightly.yml
@@ -60,12 +60,6 @@ parameters:
       DeviceName:
       TestCaseFilter: TestCategory=Smoketest-Extended
     - OS: Windows
-      Browser: Edge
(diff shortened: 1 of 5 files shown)


azure-pipelines.nightly.yml
```diff
@@ -59,12 +59,6 @@ parameters:
       DeviceType: Desktop
       DeviceName:
       TestCaseFilter: TestCategory=Smoketest-Extended
-    - OS: Windows
```
rdbartram, line 62: can you do this to the beta tests too please

npoxon: Beta tests are fine. Edge is not included, only EdgeChromium.

PR #828, opened 2020-08-27
VIH-6317 judge endpoints waiting room

valentinabennett:

https://tools.hmcts.net/jira/browse/VIH-6317

### Change description ###

display video endpoints for judge in WR

```
[ X] Yes
[ ] No
```

### DIFF
--- a/VideoWeb/VideoWeb.Contract/Responses/VideoEndpointResponse.cs
+++ b/VideoWeb/VideoWeb.Contract/Responses/VideoEndpointResponse.cs
@@ -1,0 +1,33 @@
+using System;
+using VideoWeb.Common.Models;
+
+namespace VideoWeb.Contract.Responses
+{
+    public class VideoEndpointResponse
+    {
+        /// <summary>
+        /// The endpoint id
+        /// </summary>
+        public Guid Id { get; set; }
+
+        /// <summary>
+        /// The endpoint display name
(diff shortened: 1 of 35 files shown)


VideoWeb/VideoWeb.Contract/Responses/VideoEndpointResponse.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        /// <summary>
+        /// The endpoint display name
+        /// </summary>
+        public string DisplayName { get; set; }
+
+        /// <summary>
+        /// The endpoint sip address
+        /// </summary>
+        public string SipAddress { get; set; }
```
shaed-parkar, line 21: I don't think we should expose the SIP address or PIN to the FE. These details will not be used by the app and thus should not be made public to the app

shaed-parkar (commented):
Just a concern about exposing the address and pin to the FE when this isn't used by the app and allows anyone who can see the response to see connection details for the hearing

VideoWeb/VideoWeb.Contract/Responses/VideoEndpointResponse.cs
valentinabennett: removed

shaed-parkar (approved):

PR #1025, opened 2020-11-30
VIH-6756 error page for camera mic failed

valentinabennett:

https://tools.hmcts.net/jira/browse/VIH-6756

### Change description ###

Error page for camera and mic not found or in use.

```
[ X] Yes
[ ] No
```


VideoWeb/VideoWeb/ClientApp/src/app/shared/self-test/self-test.component.ts
```diff
@@ -242,10 +242,12 @@ export class SelfTestComponent implements OnInit, OnDestroy {
             this.videoCallService.updateMicrophoneForCall(mic);
         }
         this.preferredMicrophoneStream = await this.userMediaStreamService.getStreamForMic(mic);
-        this.logger.info(`${this.loggerPrefix} Update camera and microphone selection`, {
-            cameraId: cam.deviceId,
-            microphoneId: mic.deviceId
-        });
+        if (cam && mic) {
```
shaed-parkar, line 245: we should log the device for this stream, not only log when both mic and camera are set

VideoWeb/VideoWeb/ClientApp/src/app/shared/self-test/self-test.component.ts
```diff
@@ -242,10 +242,12 @@ export class SelfTestComponent implements OnInit, OnDestroy {
...
+            this.logger.info(`${this.loggerPrefix} Update camera and microphone selection`, {
+                cameraId: cam.deviceId,
+                microphoneId: mic.deviceId
```
shaed-parkar, line 248: if the devices are null, use the existential operator to handle null values

shaed-parkar (commented):
Looks good. Just a concern about the logging. Please see my suggestions below

VideoWeb/VideoWeb/ClientApp/src/app/shared/self-test/self-test.component.ts
valentinabennett: I see, it was error on console, just tried to fix.   I removed 'If (cam && mic).

VideoWeb/VideoWeb/ClientApp/src/app/shared/error-camera-microphone/error-camera-microphone.component.ts
```diff
@@ -0,0 +1,38 @@
...
+
+@Component({
+    selector: 'app-error-camera-microphone',
+    templateUrl: './error-camera-microphone.component.html'
+})
+export class ErrorCameraMicrophoneComponent implements OnInit {
+    deviceIsInUse = false;
+
+    private readonly loggerPrefix = '[ErrorCameraMicrophonePage] -';
+    readonly ERROR_CAMERA_MIC_MESSAGE_KEY = 'vh.error.camera.mic.message';
```
shaed-parkar, line 15: Rather than creating a session storage object here to read, the error service should maintain the responsibility for saving and returning the error message rather than introducing this logic here

shaed-parkar (changes requested):
Some concern about the session storage being defined in one place to set and another to read

VideoWeb/VideoWeb/ClientApp/src/app/shared/error-camera-microphone/error-camera-microphone.component.ts
valentinabennett: It's done in the same way like it done for error.component. Just only I saved in storage flag indicated type of message on UI. And the service error maintain the responsibility for the type of error message.

VideoWeb/VideoWeb/ClientApp/src/app/services/user-media-stream.service.ts
```diff
@@ -43,26 +45,40 @@ export class UserMediaStreamService {
...
             this.stopStream(this.requestStream);
         }
-
-        this.requestStream = await this.navigator.mediaDevices.getUserMedia(this.permissionConstraints);
-        return this.requestStream;
+        try {
+            this.requestStream = await this.navigator.mediaDevices.getUserMedia(this.permissionConstraints);
+            return this.requestStream;
+        } catch (error) {
+            this.logger.error('[UserMediaStreamService] - Could not get media stream', error);
```
shaed-parkar, line 52: Please use a const for the logger prefix here, as discussed when the logger was demoed.

shaed-parkar (changes requested):
Logging prefixes should be a const variable

VideoWeb/VideoWeb/ClientApp/src/app/services/user-media-stream.service.ts
valentinabennett: Done

shaed-parkar (approved):
Nice work

PR #1042, opened 2020-12-04
VIH-6907 - audio recording alert

levibutler:
VIH-6907

### Change description ###
Alert when a problem is detected with the audio recording

image

```
[ ] Yes
[x] No
```


RuwanikaWeerakkody (dismissed):

levibutler:
@shaed-parkar 
oh that’s not right
it should display only when the hearing has started
and should be at the top of the page

ill update acs and try again

shaed-parkar (changes requested):
The alert should only be visible when the hearing is in session. The alert should also be placed at the top of the screen.

shaed-parkar (approved):
LGTM

PR #1102, opened 2021-01-05
VIH-6906 pexip error and offline error

levibutler:
VIH-6906

### Change description ###
Error handling for pexip, due to the error being the same for user offline and pexip down ive had to implement a connection status check too.

```
[ ] Yes
[X] No
```

### DIFF
--- a/VideoWeb/VideoWeb/ClientApp/src/app/services/connection-status.service.spec.ts
+++ b/VideoWeb/VideoWeb/ClientApp/src/app/services/connection-status.service.spec.ts
@@ -1,3 +1,4 @@
+import { fakeAsync, tick } from '@angular/core/testing';
 import { MockLogger } from '../testing/mocks/MockLogger';
 import { Logger } from './logging/logger-base';
 import { ConnectionStatusService } from '../services/connection-status.service';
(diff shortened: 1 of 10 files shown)


RocheL (dismissed):

VideoWeb/VideoWeb/ClientApp/src/app/services/connection-status.service.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+        // Assert
+        expect(XMLHttpRequest.prototype.open).toHaveBeenCalledTimes(2);
+        expect(XMLHttpRequest.prototype.send).toHaveBeenCalledTimes(2);
+    });
+
+    it('should stop timer if not started', () => {
+        expect(() => service.stopTimer()).not.toThrow();
+    });
+
+    it('should setup interval and fire intial check and again every 5 seconds', () => {
```
shaed-parkar, line 54: Not a big fan of a test that intentionally waits for 5+ seconds. The current test suit runs in half that time.

shaed-parkar: Would you consider the ability to change the interval and then have then use the much smaller time period?

levibutler: not waiting :) but yeah i wouldnt be either :P

shaed-parkar (approved):
LGTM

PR #1168, opened 2021-02-01
VIH-7142 Remove subscriber from network check for waiting room

jdssunil:
Remove subscriber from network check for waiting room
### Change description ###
Remove subscriber from network check for waiting room

```
[ ] Yes
[ ] No
```

### DIFF
--- a/VideoWeb/VideoWeb/ClientApp/src/app/services/connection-status.service.ts
+++ b/VideoWeb/VideoWeb/ClientApp/src/app/services/connection-status.service.ts
@@ -3,6 +3,7 @@ import { Logger } from './logging/logger-base';
 import { Observable, of, Subject } from 'rxjs';
 import { HttpClient, HttpErrorResponse } from '@angular/common/http';
 import { catchError, map } from 'rxjs/operators';
+import { LocationService } from './location.service';
 
 @Injectable({
     providedIn: 'root'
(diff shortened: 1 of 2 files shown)


VideoWeb/VideoWeb/ClientApp/src/app/services/connection-status.service.ts
```diff
@@ -17,7 +18,7 @@ export class ConnectionStatusService {
     private timer: NodeJS.Timeout;
     private pings = new Array<boolean>(this.NUMBER_OF_GOOD_PINGS_REQUIRED);
 
-    constructor(private logger: Logger, private http: HttpClient) {
+    constructor(private logger: Logger, private http: HttpClient, private locationService: LocationService) {
         this.pings.every(x => (x = true));
```
shaed-parkar, line 22: Whilst you're here, could you clean up the code smell? Thanks!

VideoWeb/VideoWeb/ClientApp/src/app/services/connection-status.service.ts
```diff
@@ -75,7 +76,10 @@ export class ConnectionStatusService {
 
         if (this.status === connectionResult) {
             this.logger.info(`${this.loggerPrefix} ${this.status ? 'Online' : 'Offline'}`);
-            this.connectionStatus.next(this.status);
+            const currentPathName = this.locationService.getCurrentPathName();
+            if (currentPathName.indexOf('waiting-room') < 0) {
```
levibutler, line 80: Hi
any reason we're doing the check in the connection service and not the subscription side?

levibutler: The connection status services responsibility is to report the status change
The subscriber's responsibility is to redirect on error (which we want to stop)
If we change in the service, any other future uses may be negatively affected by this change?

While i think this will solve the current problem, im not sure its correct?

PR #1169, opened 2021-02-01
VIH-7083 participants controller security

tremolo64:
VIH-7083

### Change description ###
Add middleware to check if participant belongs to a conference

```
[ ] Yes
[x] No
```


VideoWeb/VideoWeb/Startup.cs
```diff
@@ -33,8 +33,9 @@ public void ConfigureServices(IServiceCollection services)
             services.AddCustomTypes();
 
             services.RegisterAuthSchemes(Configuration);
-            services.AddMvc(opt => opt.Filters.Add(typeof(LoggingMiddleware))).SetCompatibilityVersion(CompatibilityVersion.Version_3_0);
-            services.AddMvc().SetCompatibilityVersion(CompatibilityVersion.Version_3_0).AddFluentValidation();
+            var compatibilityVersion = CompatibilityVersion.Version_3_0;
```
shaed-parkar, line 36: If this is a recommendation from resharper it may be worth having rules added to the repo so different users do not get different recommendations.

shaed-parkar (approved):
Nice work!

shaed-parkar (commented):
The Authorize attribute may be suitable in a few places due to the fact only certain roles will use it

VideoWeb/VideoWeb/Controllers/ParticipantsController.cs
```diff
@@ -142,6 +144,7 @@ public async Task<IActionResult> GetIndependentTestCallResultAsync(Guid particip
             }
         }
 
+        [ServiceFilter(typeof(CheckParticipantCanAccessConferenceAttribute))]
         [HttpGet("{conferenceId}/participant/{participantId}/heartbeatrecent")]
         [SwaggerOperation(OperationId = "GetHeartbeatDataForParticipant")]
         [ProducesResponseType(typeof(ParticipantHeartbeatResponse[]), (int)HttpStatusCode.OK)]
```
shaed-parkar, line 150: `[Authorize(AppRoles.VhOfficerRole)]` might be better there rather than the service filter since only VHOfficers use this endpoint

VideoWeb/VideoWeb/Startup.cs
tremolo64: Removed the SetCompatibilityVersion method as it is a no-op for ASP.NET Core 3.0 apps. Calling SetCompatibilityVersion with any value of CompatibilityVersion has no impact on the application.

See: https://docs.microsoft.com/en-us/aspnet/core/mvc/compatibility-version?view=aspnetcore-3.1

VideoWeb/VideoWeb/Controllers/ParticipantsController.cs
tremolo64: The filter is applied globally, and I think this particular endpoint can used by any participant.

tremolo64: I've added the [Authorize] filter now.

tremolo64:
@shaed-parkar I've changed the tests to use ClaimsIdentityBuilder now

shaed-parkar (approved):

PR #1187, opened 2021-02-11
VIH-7118 Buttons layout for wr judge

valentinabennett:
https://tools.hmcts.net/jira/browse/VIH-7118

### Change description ###
 The button consultation made disabled until the connection is established and fix the layout of the buttons for iPad portrait view.

```
[X ] Yes
[ ] No
```

### DIFF
--- a/VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/judge-waiting-room/judge-waiting-room.component.scss
+++ b/VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/judge-waiting-room/judge-waiting-room.component.scss
@@ -40,6 +40,7 @@
 }
 
 .vh-btn-text {
+    font-size: 24px;
 }
 
 @media only screen and (max-width: 1024px) {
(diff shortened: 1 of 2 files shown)


VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/judge-waiting-room/judge-waiting-room.component.scss
```diff
@@ -38,3 +38,12 @@
   max-height: 85vh;
   z-index: 100;
 }
+
+.vh-btn-text {
+}
+
+@media screen and (max-width: 1024px) {
```
shaed-parkar, line 45: I think you need to use screen only to make sure normal devices aren't affected by this change. This page has a list of combinations that affect tablets only. I hope this helps.
[css device queries](https://css-tricks.com/snippets/css/media-queries-for-standard-devices/#ipad-queries)

shaed-parkar (commented):
Maybe we can create some functions or helpers in scss to determine if the browser is an ipad

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/judge-waiting-room/judge-waiting-room.component.scss
valentinabennett: I added 'only'
The fix should be ok for normal devices as it had not the correct layout if you resize the view, now it's should be better.

shaed-parkar (approved):
Please demo with the QA team before merging.

PR #1257, opened 2021-03-02
VIH-7149 interpreter-interpretee displayed together in the UI

williambowshersolirius:
https://tools.hmcts.net/jira/browse/VIH-7149

### Change description ###
Changed Participant lists in the UI so that if an Individual has an interpreter, then the interpreter is the next participant displayed in the list and both participants have the translator icon displayed next to their names

```
[ ] Yes
[x ] No
```


VideoWeb/VideoWeb/ClientApp/src/app/testing/mocks/data/conference-test-data.ts
```diff
@@ -324,7 +326,113 @@ export class ConferenceTestData {
 
     getListOfParticipantDetails(): ParticipantResponseVho[] {
         const participants: ParticipantResponseVho[] = [];
+
```
shaed-parkar, line 329: This looks like a copy and paste of the previous method but for `ParticipantResponseVho`. And the following method returns a list of `ParticipantResponseVho`. Could you explain this change? I'm not able to follow

VideoWeb/VideoWeb/Program.cs
```diff
@@ -1,5 +1,6 @@
 using System.IO;
 using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.Configuration;
```
shaed-parkar, line 3: Added a using statement, but no other line change?

VideoWeb/VideoWeb/ClientApp/src/app/testing/mocks/data/conference-test-data.ts
williambowshersolirius: This was due to the fact I wanted to set the linked_participants, and I have a flag hasInterpreterLink on the ParticipantResponse(VHO) class that needs to get set. Plus I was trying to limit the impact of changing test data as getListOfParticipantDetails seems to be used in the conference setup

PR #1274, opened 2021-03-04
VIH-7201 update eventhub clients to publish mute and hand raised for vmr

shaed-parkar:

VIH-7201 

##### **Part 1 - update the clients** #####
Participants in a VMR do not receive callbacks like other participants, nor can they raise their hand like other participants. The `EventHubClient` is extended to provide this to room participants

```
[ ] Yes
[x] No
```

### DIFF
--- a/VideoWeb/VideoWeb.EventHub/Hub/EventHubClient.cs
+++ b/VideoWeb/VideoWeb.EventHub/Hub/EventHubClient.cs
@@ -41,7 +41,7 @@ namespace VideoWeb.EventHub.Hub
         public override async Task OnConnectedAsync()
         {
             var userName = await GetObfuscatedUsernameAsync(Context.User.Identity.Name);
-            _logger.LogTrace("Connected to event hub server-side: {userName}", userName);
+            _logger.LogTrace("Connected to event hub server-side: {Username}", userName);
             var isAdmin = IsSenderAdmin();
 
             await AddUserToUserGroup(isAdmin);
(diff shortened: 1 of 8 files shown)


VideoWeb/VideoWeb.EventHub/Hub/EventHubClient.cs
```diff
@@ -404,6 +408,77 @@ await Clients.Group(judge.Username.ToLowerInvariant())
...
+                        .Group(linkedParticipant.Username.ToLowerInvariant())
+                        .ParticipantHandRaiseMessage(participantId, conferenceId, isRaised)).ToArray());
+                _logger.LogTrace(
+                    "Participant hand status updated: Participant Id: {ParticipantId} | Conference Id: {ConferenceId} to {IsHandRaised}",
+                    participantId, conferenceId, isRaised);
+
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(ex,
```
levibutler, line 476: i dont agree with catching just to add logging but i see its done for the other methods so letting it slide

shaed-parkar: I think we've had problems int the past when an unhandled exception kills the connection

VideoWeb/VideoWeb/ClientApp/src/app/services/events.service.ts
```diff
@@ -210,6 +212,24 @@ export class EventsService {
             }
         );
 
+        this.connection.on('ParticipantRemoteMuteMessage', (participantId: string, conferenceId: string, isRemoteMuted: boolean) => {
+            this.logger.debug('[EventsService] - Participant Remote mute status change received: ', {
+                participantId,
+                conferenceId,
+                isRemoteMuted
+            });
+            this.participantRemoteMuteStatusSubject.next(isRemoteMuted);
```
levibutler, line 221: we dont need the participant or conference because the even will only come through to the relvant participants

levibutler (approved):
LGTM

PR #1334, opened 2021-03-18
VIH-7482 video hearing bug fixes for tile position and muted video element

shaed-parkar:

VIH-7482

### Change description ###

* ensure tile positions are unique
* make sure video stream isn't muted when browser is refreshed mid-hearing

```
[ ] Yes
[x] No
```

### DIFF
--- a/VideoWeb/VideoWeb/Mappings/SharedParticipantRoomMapper.cs
+++ b/VideoWeb/VideoWeb/Mappings/SharedParticipantRoomMapper.cs
@@ -13,8 +13,9 @@ namespace VideoWeb.Mappings
         {
             var node = sharedRoom.PexipNode.Replace("https://", string.Empty);
             var tilePosition = int.Parse(new string(sharedRoom.Label.Where(char.IsDigit).ToArray()));
+            var randomNumber = GetRandomNumberAsString();
             var tilePrefix = isWitness ? "W" : "T";
-            var tileDisplayName = $"{tilePrefix}{200+tilePosition};{participant.DisplayName};{participant.Id}";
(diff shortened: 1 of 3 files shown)


RuwanikaWeerakkody (changes requested):
Requesting a change

VideoWeb/VideoWeb/Mappings/SharedParticipantRoomMapper.cs
```diff
@@ -13,8 +13,9 @@ public SharedParticipantRoom Map(SharedParticipantRoomResponse sharedRoom, Parti
         {
             var node = sharedRoom.PexipNode.Replace("https://", string.Empty);
             var tilePosition = int.Parse(new string(sharedRoom.Label.Where(char.IsDigit).ToArray()));
+            var randomNumber = GetRandomNumberAsString();
             var tilePrefix = isWitness ? "W" : "T";
-            var tileDisplayName = $"{tilePrefix}{200+tilePosition};{participant.DisplayName};{participant.Id}";
+            var tileDisplayName = $"{tilePrefix}{200+tilePosition}{randomNumber};{participant.DisplayName};{participant.Id}";
```
RuwanikaWeerakkody, line 18: I don't think this is the correct fix Shaed. The issue was both interpreter and Witness had tile numbers starting with W. Only witness should have a W. Interpreter should still get a T.

shaed-parkar: I originally had interpreters for witness starting with a T and they would get dragged into a hearing like everyone else on start. The conversation looks like there was an issue with both participants having a tile position of W201

RuwanikaWeerakkody (approved):

PR #1360, opened 2021-03-29
VIH-7314 - event hub auth to use access token and subscribe

levibutler:
VIH-7314

### Change description ###
event hub auth to use access token and subscribe

```
[ ] Yes
[x] No
```


shaed-parkar (approved):
LGTM

PR #1366, opened 2021-03-31
VIH-7548 handle ims for participant when unread array is empty

shaed-parkar:

VIH-7548

### Change description ###

* unread components need to handle incoming messages differently
handle ims for a participant when the unread array is empty

```
[ ] Yes
[x] No
```


kevin-hmcts (approved):
Looks good to me.

PR #1371, opened 2021-04-08
VIH-7091 duplicate IM messages

robertbennett1998:
https://tools.hmcts.net/jira/browse/VIH-7091

### Change description ###

- Added a check to ensure that the SignalR event handlers in events.service are only registered once.
- Refactored events service so that it only contains the observables for recieve messages and logic for send messages.
- Moved the hub connection logic from events service into event hub service.
- Events hub service manages the hub connection and handles all reconnection cases such that no other place in the code needs to call start/stop with the exception of the initial start in events service.

```
[ ] Yes
[x] No
```


shaed-parkar (approved):
LGTM

VideoWeb/VideoWeb/ClientApp/src/app/app.component.spec.ts
```diff
@@ -110,15 +108,6 @@ describe('AppComponent', () => {
         expect(connectionStatusServiceSpy.start).toHaveBeenCalled();
     }));
 
-    it('should start event service if authenticated oninit', fakeAsync(() => {
```
shaed-parkar, line 113: When do you propose to start the connection to EventHub if not once authenticated?

robertbennett1998: The constructor of EventsHubService now does the building of the HubConnection which raises an event when it is set up which is picked up by EventsService; EventsService then starts the connection. The service is never created until after the user has logged in and has been redirected from the MS login page, so it cannot start until the user is authenticated. The start method also checks if the user is authenticated before actually starting the HubConnection.

VideoWeb/VideoWeb/ClientApp/src/app/services/events-hub.service.ts
```diff
@@ -0,0 +1,213 @@
...
+                this.start();
+            });
+        } else {
+            this.logger.info(`[EventsService] - Failed to connect too many times (#${this.reconnectionAttempt}), going to service error`);
+            this.errorService.goToServiceError('Your connection was lost');
+
+            // Only subscibe to the first emitted event where the reconnection was successful
+            this.errorService.onUserTriggeredReconnect
+                .pipe(
+                    filter(reconnectionSuccessful => reconnectionSuccessful),
```
phillipjchambers, line 147: Little shorthand: `filter(Boolean)`
It'll pass the value to Boolean's ctor, which returns true/false based on the constructed bool

phillipjchambers (approved):

kevin-hmcts (approved):

shaed-parkar (approved):
Looks good please run a hearing and consultation before merging in please

PR #1375, opened 2021-04-12
VIH-7553  Video web login logout config issue

ZJianWang:

VIH-7553-VIH-Video-Web|Login/out config issues

### Change description ###

Add auth.service to make login process to when users complete checkAuth we check isAuthenticated$

```
[ ] Yes
[x ] No
```


VideoWeb/VideoWeb/ClientApp/src/app/app.component.ts
```diff
@@ -90,7 +90,11 @@ export class AppComponent implements OnInit, OnDestroy {
     private async attemptRetrieveProfile(loggedIn: boolean) {
         if (!loggedIn) {
             const currentUrl = this.locationService.getCurrentUrl();
-            this.router.navigate([`/${pageUrls.IdpSelection}`], { queryParams: { returnUrl: currentUrl } });
+            if (currentUrl.indexOf(pageUrls.Logout) > -1) {
+                this.router.navigate([`/${pageUrls.Logout}`]);
```
phillipjchambers, line 94: hey @ZJianWang - can you explain this change, pls?

VideoWeb/VideoWeb/ClientApp/src/app/services/security/auth.service.ts
```diff
@@ -0,0 +1,30 @@
+import { Injectable } from '@angular/core';
```
phillipjchambers, line 1: @ZJianWang why is this additonal service required? It looks like it's just a sort of proxy for the oidcSecurityService.

VideoWeb/VideoWeb/ClientApp/src/app/services/security/auth.service.ts
```diff
@@ -0,0 +1,30 @@
...
+
+    public get isAuthenticated$(): Observable<boolean> {
+        return this.checkAuthCompleted$.pipe(
+            first(),
+            switchMap((_) => this.oidcSecurityService.isAuthenticated$)
+        );
+    }
+
+    public checkAuth(): Observable<boolean> {
+        return this.oidcSecurityService.checkAuth().pipe(tap((_) => this.checkAuthCompleted$.next()));
```
phillipjchambers, line 20: Looks like this just emits the checkAuthCompleted$ subject when checkAuth emits. Could we just return a reference to the checkAuth observable?

phillipjchambers: Instead of adding another observable to subscibe to, I mean.

VideoWeb/VideoWeb/ClientApp/src/app/app.component.ts
ZJianWang: looks this change need to revert I saw Nick already had a fixed for logout issue

VideoWeb/VideoWeb/ClientApp/src/app/services/security/auth.service.ts
ZJianWang: This is to make ignore values from isAuthenticated$ before checkAuth() emits a value, I don't have idead to just return a reference, I think create an another observable to subscibe is necessary

VideoWeb/VideoWeb/ClientApp/src/app/services/security/auth.service.ts
ZJianWang: OK I will try to this without creating a new service

ZJianWang: looks like need additonal service to share checkAuthCompleted between app.component.ts and auth.guard.tes

shaed-parkar: but why? what does this additional layer provide which you can't achieve with just `oidcSecurityService`?

shaed-parkar (changes requested):
The auth service adds a layer of abstraction we do not need. There seems to be a lot of change here for something quite trivial. If you have a look at the library repo [here](https://github.com/damienbod/angular-auth-oidc-client) you'll see the issue is the cause of the initial value of a Behaviour Subject.

A single line with `Skip` would suffice in the Auth Guard

PR #1389, opened 2021-04-21
VIH-7553 login issue fixing

ZJianWang:
VIH-7553

### Change description ###
Add a delay 1 second to wait for the checkauth complete

```
[ ] Yes
[x] No
```


shaed-parkar (changes requested):
Hi, as discussed the issue lies in the fact that the auth guard is applied in the default route. This should not be merged in as this does not solve the fundamental issue. 

Please see this document again [Guard with OIDC Client](https://github.com/damienbod/angular-auth-oidc-client/blob/main/docs/guards.md)

PR #1418, opened 2021-05-18
VIH-7538 Added Role Claims to JWT Kinly

alberto-hortiguela:
https://tools.hmcts.net/jira/browse/VIH-7538

### Change description ###
Added role claim to Kinly JWT. ( Kinly will use the claim at their end to authorized consuming their API endpoints)

```
[ x ] Yes
[ ] No
```


alberto-hortiguela:
We do not need to implement role claims on the token. There are two custom tokens we generate using keys with symmetric encryption. These keys are shared keys between us and Kinly. There are two instances where the custom tokens are used.
Heartbeat calls from video web frontend app
Making calls to Kinly API from Video API
The keys that we use to generate for the above should be different. These keys are different on preprod and prod but they are same on lower environments. I believe this is what is allowing to make calls to Kinly API successfully using heartbeat token. The fix for this to generate a new Key and use it for heartbeat tokens for lower environments. By the way, the keys are configurable on app settings.

PR #1443, opened 2021-06-09
VIH-7787 Participant status not updating in command center

ZJianWang:

VIH-7787

### Change description ###

Do some test on SignalR, don't merge

```
[ ] Yes
[x] No
```

### DIFF
--- a/VideoWeb/VideoWeb.Common/Models/Conference.cs
+++ b/VideoWeb/VideoWeb.Common/Models/Conference.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using VideoApi.Contract.Enums;
 
 namespace VideoWeb.Common.Models
 {
@@ -19,6 +20,7 @@ namespace VideoWeb.Common.Models
         public List<Endpoint> Endpoints { get; set; }
         public List<CivilianRoom> CivilianRooms { get; set; }
         public string HearingVenueName { get; set; }
+        public ConferenceState CurrentStatus { get; set; }
 
         public Participant GetJudge()
         {
--- a/VideoWeb/VideoWeb.Common/Models/Conference.cs
+++ b/VideoWeb/VideoWeb.Common/Models/Conference.cs
@@ -54,9 +54,5 @@ namespace VideoWeb.Common.Models
 
             return room;
         }
-        public bool IsConferenceInSession()
-        {
-            return CurrentStatus == ConferenceState.InSession;
-        }
     }
 }
--- a/VideoWeb/VideoWeb.Common/Models/Conference.cs
+++ b/VideoWeb/VideoWeb.Common/Models/Conference.cs
@@ -54,5 +54,9 @@ namespace VideoWeb.Common.Models
 
             return room;
         }
+        public bool IsConferenceInSession()
+        {
+            return CurrentStatus == ConferenceState.InSession;
+        }
     }
 }
--- a/VideoWeb/VideoWeb.EventHub/Models/CallbackEvent.cs
+++ b/VideoWeb/VideoWeb.EventHub/Models/CallbackEvent.cs
@@ -16,5 +16,8 @@ namespace VideoWeb.EventHub.Models
         public string TransferTo { get; set; }
 
         public string Reason { get; set; }
+        public bool IsParticipantInVmr { get; set; }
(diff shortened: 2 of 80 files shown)


VideoWeb/VideoWeb.Common/Models/Conference.cs
```diff
@@ -52,5 +54,9 @@ private CivilianRoom GetOrCreateCivilianRoom(long roomId)
 
             return room;
         }
+        public bool IsConferenceInSession()
+        {
+            return CurrentStatus == ConferenceState.InSession;
+        }
```
FarronF, line 60: Could you chanage this to be a get property?

VideoWeb/VideoWeb/Extensions/ConferenceEventRequestExtensions.cs
```diff
@@ -24,43 +32,51 @@ public static bool IsParticipantAndVmrEvent(this ConferenceEventRequest request)
...
         }
+
+        public static void UpdateEventsTypeForVmrParticipants(this ConferenceEventRequest request,
+            Conference conference)
+        {
+            if (request.ParticipantRoomId.IsNullOrEmpty() || !conference.IsConferenceInSession() ||
+                !request.IsParticipantInVmr(conference)) return;
+            request.EventType = EventType.RoomParticipantTransfer;
+            request.TransferTo = nameof(RoomType.HearingRoom);
+        }
```
FarronF, line 80: I don't think we should change the event type of a request like this. Could we get these details in the JoinedEventHandler?

ZJianWang: No, there is nothing we can use from the JoinedEvent,

FarronF: I see. Could this logic be part of CreateEventsForParticipantsInRoom instead then? We could generate 2 events for a user that's joining if the conference is in session instead.

FarronF: I misread the if statement that calls CreateEventsForParticipantsInRoom. This wouldn't work. I still don't think we should change the request event type and should either have multiple events or have the handler be able to determine where the vmr is and set the participant status based on that.

I also don't think this will work if the JOH's are in a consultation.

VideoWeb/VideoWeb.Common/Models/Conference.cs
```diff
@@ -52,5 +54,9 @@ private CivilianRoom GetOrCreateCivilianRoom(long roomId)
 
             return room;
         }
+        public bool IsConferenceInSession()
```
FarronF, line 57:
```suggestion
        public boolean IsInSession => CurrentStatus == ConferenceState.InSession;
```

VideoWeb/VideoWeb.EventHub/Models/CallbackEvent.cs
```diff
@@ -16,5 +19,10 @@ public class CallbackEvent
         public string TransferTo { get; set; }
 
         public string Reason { get; set; }
+        public bool IsParticipantInVmr { get; set; }
+        public ConferenceState ConferenceStatus { get; set; }
+        public IEnumerable<Participant> OtherParticipantsInVmr { get; set; }
```
FarronF, line 24: I don't think we need this as it's only used to set up IsOtherParticipantsInConsultationRoom

ZJianWang: Hi Farron, IsParticipantInVmr  and ConferenceStatus  are using in JoinedEventhandler, I will remove OtherParticipantsInVmr

VideoWeb/VideoWeb.Common/Models/Conference.cs
ZJianWang: This is no need any more I will remove this line

VideoWeb/VideoWeb.EventHub/Models/CallbackEvent.cs
FarronF: Yeah I actually just meant that one, I must have marked the others by mistake.

FarronF (approved):

PR #1445, opened 2021-06-14
VIH-7827 added lifecycle hook to update component

omio-ma:
https://tools.hmcts.net/jira/browse/VIH-7827

### Change description ###
- Added on change lifecycle method for the component to update the text when language is changed. 

```
[ ] Yes
[x] No
```

### DIFF
--- a/VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/select-hearing-layout/select-hearing-layout.component.spec.ts
+++ b/VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/select-hearing-layout/select-hearing-layout.component.spec.ts
@@ -40,7 +40,7 @@ describe('SelectHearingLayoutComponent', () => {
         expect(component.currentButtonContentKey).toBe(buttonContentKeyWhenOpen);
     });
 
-    it('should call translate service to update accordion button when language changes', () => {
+    it('should show translated text on open accordion button', () => {
         const expectedTranslatedContentForButton = 'this is translated for open all button';
         component.currentButtonContentKey = buttonContentKeyWhenOpen;
         translateServiceSpy.instant
@@ -51,7 +51,7 @@ describe('SelectHearingLayoutComponent', () => {
         expect(component.accordionOpenAllElement.innerHTML).toContain(expectedTranslatedContentForButton);
     });
 
-    it('should call translate service to update accordion header when language changes', () => {
+    it('should show translated text on open/close toggle button', () => {
         const expectedTranslatedContentForHeader = 'this is translated for the accordion header';
         translateServiceSpy.instant
             .withArgs('select-hearing-layout.choose-hearing-layout')
--- a/VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/select-hearing-layout/select-hearing-layout.component.spec.ts
+++ b/VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/select-hearing-layout/select-hearing-layout.component.spec.ts
@@ -14,6 +14,8 @@ describe('SelectHearingLayoutComponent', () => {
     const translateService = translateServiceSpy;
     const headingButton = document.createElement('button');
     const textButton = document.createElement('button');
+    const buttonContentKeyWhenOpen = 'open-all';
+    const buttonContentKeyWhenClosed = 'close-all';
 
     beforeEach(() => {
         conference = new ConferenceTestData().getConferenceDetailNow();
@@ -21,6 +23,11 @@ describe('SelectHearingLayoutComponent', () => {
         component.conference = conference;
         textButton.innerHTML = 'Open all';
         document.getElementById = jasmine.createSpy('accordion-choose-layout-heading').and.returnValue(headingButton);
+        document.getElementsByClassName = jasmine.createSpy('govuk-accordion__section-button').and.returnValue({
+            item() {
+                return headingButton;
+            }
+        });
         document.getElementsByClassName = jasmine.createSpy('govuk-accordion__open-all').and.returnValue({
             item() {
                 return textButton;
@@ -35,17 +42,30 @@ describe('SelectHearingLayoutComponent', () => {
         videoCallService.getPreferredLayout.and.returnValue(layout);
         component.ngOnInit();
         expect(component.selectedLayout).toBe(layout);
+        expect(component.currentButtonContentKey).toBe(buttonContentKeyWhenOpen);
     });
 
-    it('should call translate service to update hearing layout option when language changes', () => {
-        component.currentText = textButton.innerHTML.toLowerCase().split(' ').join('-');
-        component.accordionOpenAllElement = document.getElementsByClassName('govuk-accordion__open-all').item(0) as HTMLButtonElement;
+    it('should call translate service to update accordion button when language changes', () => {
+        const expectedTranslatedContentForButton = 'this is translated for open all button';
+        component.currentButtonContentKey = buttonContentKeyWhenOpen;
+        translateServiceSpy.instant
+            .withArgs(`select-hearing-layout.${component.currentButtonContentKey}`)
+            .and.returnValue(expectedTranslatedContentForButton);
         component.ngOnInit();
         onLangChangeSpy.emit({ lang: 'tl' } as LangChangeEvent);
-        expect(component.currentText).toBe('open-all');
-        expect(translateServiceSpy.instant).toHaveBeenCalledWith('select-hearing-layout.choose-hearing-layout');
-        expect(translateServiceSpy.instant).toHaveBeenCalledWith(`select-hearing-layout.${component.currentText}`);
-        expect(component.accordionOpenAllElement.innerHTML).toContain(textButton.innerHTML);
+        expect(component.accordionOpenAllElement.innerHTML).toContain(expectedTranslatedContentForButton);
+        expect(component.currentButtonContentKey).toBe(buttonContentKeyWhenOpen);
+    });
+
+    it('should call translate service to update accordion header when language changes', () => {
+        const expectedTranslatedContentForHeader = 'this is translated for the accordion header';
+        component.currentButtonContentKey = buttonContentKeyWhenOpen;
+        translateServiceSpy.instant
+            .withArgs('select-hearing-layout.choose-hearing-layout')
+            .and.returnValue(expectedTranslatedContentForHeader);
+        component.ngOnInit();
+        onLangChangeSpy.emit({ lang: 'tl' } as LangChangeEvent);
+        expect(headingButton.innerHTML).toContain(expectedTranslatedContentForHeader);
     });
 
     it('should translate button text on text click', () => {
@@ -53,6 +73,7 @@ describe('SelectHearingLayoutComponent', () => {
         textButton.innerHTML = 'Close all';
         textButton.click();
         expect(textButton.innerHTML).toBe('<span>select-hearing-layout.close-all</span>');
+        expect(component.currentButtonContentKey).toBe(buttonContentKeyWhenClosed);
     });
 
     it('should translate button text on header click', () => {
@@ -60,6 +81,7 @@ describe('SelectHearingLayoutComponent', () => {
         textButton.innerHTML = 'Close all';
         headingButton.click();
         expect(textButton.innerHTML).toBe('<span>select-hearing-layout.close-all</span>');
+        expect(component.currentButtonContentKey).toBe(buttonContentKeyWhenClosed);
     });
 
     it('should translate button text on after time for header click', fakeAsync(() => {
@@ -68,6 +90,7 @@ describe('SelectHearingLayoutComponent', () => {
         textButton.innerHTML = 'Close all';
         tick(10);
         expect(textButton.innerHTML).toBe('<span>select-hearing-layout.close-all</span>');
+        expect(component.currentButtonContentKey).toBe(buttonContentKeyWhenClosed);
     }));
 
     it('should use recommended layout when cached layout preference is empty on init', () => {
--- a/VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/select-hearing-layout/select-hearing-layout.component.spec.ts
+++ b/VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/select-hearing-layout/select-hearing-layout.component.spec.ts
@@ -46,17 +46,25 @@ describe('SelectHearingLayoutComponent', () => {
             component.currentButtonContentKey = buttonContentKeyWhenOpen;
             translateServiceSpy.instant
                 .withArgs(`select-hearing-layout.${component.currentButtonContentKey}`)
-                .and.returnValue(expectedTranslatedContentForButton);
+                .and.returnValue('initial content');
             component.ngOnInit();
+            translateServiceSpy.instant
+                .withArgs(`select-hearing-layout.${component.currentButtonContentKey}`)
+                .and.returnValue(expectedTranslatedContentForButton);
+            translateServiceSpy.onLangChange.emit({ lang: 'tl' } as LangChangeEvent);
             expect(component.accordionOpenAllElement.innerHTML).toContain(expectedTranslatedContentForButton);
         });
 
         it('should show translated text on open/close toggle button', () => {
             const expectedTranslatedContentForHeader = 'this is translated for the accordion header';
             translateServiceSpy.instant
+                .withArgs(`select-hearing-layout.${component.currentButtonContentKey}`)
+                .and.returnValue('initial content');
+            component.ngOnInit();
+            translateServiceSpy.instant
                 .withArgs('select-hearing-layout.choose-hearing-layout')
                 .and.returnValue(expectedTranslatedContentForHeader);
-            component.ngOnInit();
+            translateServiceSpy.onLangChange.emit({ lang: 'tl' } as LangChangeEvent);
             expect(headingButton.innerHTML).toContain(expectedTranslatedContentForHeader);
         });
     });
--- a/VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/select-hearing-layout/select-hearing-layout.component.spec.ts
+++ b/VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/select-hearing-layout/select-hearing-layout.component.spec.ts
@@ -40,25 +40,25 @@ describe('SelectHearingLayoutComponent', () => {
         expect(component.currentButtonContentKey).toBe(buttonContentKeyWhenOpen);
     });
 
-    it('should show translated text on open accordion button', () => {
-        const expectedTranslatedContentForButton = 'this is translated for open all button';
-        component.currentButtonContentKey = buttonContentKeyWhenOpen;
-        translateServiceSpy.instant
-            .withArgs(`select-hearing-layout.${component.currentButtonContentKey}`)
-            .and.returnValue(expectedTranslatedContentForButton);
-        component.ngOnInit();
-        onLangChangeSpy.emit({ lang: 'tl' } as LangChangeEvent);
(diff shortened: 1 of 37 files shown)


VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/select-hearing-layout/select-hearing-layout.component.ts
```diff
@@ -30,6 +31,17 @@ export class SelectHearingLayoutComponent implements OnInit {
         this.setAccordionText({} as MouseEvent);
     }
 
+    ngOnChanges(): void {
```
robertbennett1998, line 34: Should this not be in ngOnInit and then unsubscribed in ngOnDestroy. Won't subscribing here set up multiple listeners?

omio-ma: Do you mean to move the onlangchange event into the ngoninit function? I didn't test that way but wouldn't that not update the page without a hard reload?

robertbennett1998: Yeah, no it shouldn't as you are subscribing to an event and then just updating some values.

omio-ma: removed the lifecycle method, I thought it gets called whenever the state changes in the app, but then I realized we actually dont have a global state instead we use local/session storage.

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/select-hearing-layout/select-hearing-layout.component.spec.ts
```diff
@@ -36,6 +37,17 @@ describe('SelectHearingLayoutComponent', () => {
         expect(component.selectedLayout).toBe(layout);
     });
 
+    it('should call translate service to update hearing layout option when language changes', () => {
+        component.currentText = textButton.innerHTML.toLowerCase().split(' ').join('-');
+        component.accordionOpenAllElement = document.getElementsByClassName('govuk-accordion__open-all').item(0) as HTMLButtonElement;
```
FarronF, line 42: Shouldn't this be done within the ngOnInit code rather than directly here?

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/select-hearing-layout/select-hearing-layout.component.spec.ts
```diff
@@ -36,6 +37,17 @@ describe('SelectHearingLayoutComponent', () => {
...
+        component.ngOnInit();
+        onLangChangeSpy.emit({ lang: 'tl' } as LangChangeEvent);
+        expect(component.currentText).toBe('open-all');
+        expect(translateServiceSpy.instant).toHaveBeenCalledWith('select-hearing-layout.choose-hearing-layout');
+        expect(translateServiceSpy.instant).toHaveBeenCalledWith(`select-hearing-layout.${component.currentText}`);
```
FarronF, line 47: I would make translateServiceSpy.instant return a specific value we can test for when called by this string.

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/select-hearing-layout/select-hearing-layout.component.spec.ts
omio-ma: I didn't want to hard code some string here, and rather use the textbutton defined at the top of the test file, or are you referring to the accordionOpenAllElement?

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/select-hearing-layout/select-hearing-layout.component.spec.ts
```diff
@@ -36,6 +37,17 @@ describe('SelectHearingLayoutComponent', () => {
         expect(component.selectedLayout).toBe(layout);
     });
 
+    it('should call translate service to update hearing layout option when language changes', () => {
+        component.currentText = textButton.innerHTML.toLowerCase().split(' ').join('-');
```
FarronF, line 41: I think you should setup your own values for this test rather than pulling from textButton.

omio-ma: yea sure I can do that

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/select-hearing-layout/select-hearing-layout.component.spec.ts
omio-ma: ok cool, I will attempt to do that.

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/select-hearing-layout/select-hearing-layout.component.ts
```diff
@@ -39,8 +49,8 @@ export class SelectHearingLayoutComponent implements OnInit {
         const text = this.accordionOpenAllElement.innerHTML;
         if (!text.startsWith('<')) {
             const originalText = text.split('<')[0];
-            const currentText = originalText.toLowerCase().split(' ').join('-').trim();
-            const translated = `<span>${this.translateService.instant(`select-hearing-layout.${currentText}`)}</span>`;
+            this.currentText = originalText.toLowerCase().split(' ').join('-').trim();
+            const translated = `<span>${this.translateService.instant(`select-hearing-layout.${this.currentText}`)}</span>`;
```
FarronF, line 53: We can test that currentText is set correctly now. Maybe make it more clear what this does, this is because we have no control over that button right? So it will always be either open-all or close-all.

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/select-hearing-layout/select-hearing-layout.component.spec.ts
omio-ma: oh actually, I dont need that part with accordionOpenAllElement anymore, as I have moved the code back into ngoninit, previously it was in ngonchanges and I was testing it in an isolated way (ie, calling the function directly). I think I should be able to remove that

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/select-hearing-layout/select-hearing-layout.component.ts
FarronF: Could some of this be another function since we do a very similar process in both this function and the subscribe?

omio-ma: yea I can add assertion to check what the expected current text is in all the relevant tests, shall I do that?

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/select-hearing-layout/select-hearing-layout.component.spec.ts
FarronF: https://stackoverflow.com/questions/16198353/any-way-to-modify-jasmine-spies-based-on-arguments

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/select-hearing-layout/select-hearing-layout.component.ts
omio-ma: regarding re using that if block, the logic appears to be specific to the lifecycle of the component itself. Initially the innerhtml doesn't appear to start with a "<" but in my case the accordion will have that and also, even if I were to use the the same logic it appears after using "split" on the innerHTML, the part of text that I require to translate no longer is in the the array index of 0. So instead I used innerText in ngoninit. Might be easier to show it rather than explaining here

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/select-hearing-layout/select-hearing-layout.component.spec.ts
```diff
@@ -13,13 +14,20 @@ describe('SelectHearingLayoutComponent', () => {
...
         conference = new ConferenceTestData().getConferenceDetailNow();
         component = new SelectHearingLayoutComponent(videoCallService, translateService);
         component.conference = conference;
         textButton.innerHTML = 'Open all';
         document.getElementById = jasmine.createSpy('accordion-choose-layout-heading').and.returnValue(headingButton);
+        document.getElementsByClassName = jasmine.createSpy('govuk-accordion__section-button').and.returnValue({
+            item() {
+                return headingButton;
+            }
+        });
```
FarronF, line 30: I'm not really sure what's going on here, even with the old code. Do lines 26-30 not get overwritten by 31-35 or can they co-exist because of the different names?

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/select-hearing-layout/select-hearing-layout.component.spec.ts
```diff
@@ -34,20 +42,46 @@ describe('SelectHearingLayoutComponent', () => {
...
+            .and.returnValue(expectedTranslatedContentForButton);
+        component.ngOnInit();
+        onLangChangeSpy.emit({ lang: 'tl' } as LangChangeEvent);
+        expect(component.accordionOpenAllElement.innerHTML).toContain(expectedTranslatedContentForButton);
+        expect(component.currentButtonContentKey).toBe(buttonContentKeyWhenOpen);
+    });
+
+    it('should call translate service to update accordion header when language changes', () => {
+        const expectedTranslatedContentForHeader = 'this is translated for the accordion header';
+        component.currentButtonContentKey = buttonContentKeyWhenOpen;
```
FarronF, line 62: This line isn't needed for this test is it?

omio-ma: removed it, good spot

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/select-hearing-layout/select-hearing-layout.component.spec.ts
omio-ma: strange, I thought this is the way of mocking the dom document with expected argument. but I am assuming in jasmine the mocks should have been something like:
`spyOn(document, "getElementsByClassName ").withArgs(`govuk-accordion__section-button`).and.returnValue("whatever")`
Anyways, i have removed the line I have added its not required I think I had it from earlier so once again good spot! :D

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/select-hearing-layout/select-hearing-layout.component.spec.ts
```diff
@@ -34,20 +37,45 @@ describe('SelectHearingLayoutComponent', () => {
...
+    });
+
+    it('should call translate service to update accordion header when language changes', () => {
+        const expectedTranslatedContentForHeader = 'this is translated for the accordion header';
+        translateServiceSpy.instant
+            .withArgs('select-hearing-layout.choose-hearing-layout')
+            .and.returnValue(expectedTranslatedContentForHeader);
+        component.ngOnInit();
+        onLangChangeSpy.emit({ lang: 'tl' } as LangChangeEvent);
+        expect(headingButton.innerHTML).toContain(expectedTranslatedContentForHeader);
```
AWildBugAppeared, line 62: **TLDR: Tests should not test things that are outside the scope of the test name.**

Minor change. Can the names of these tests be changed?

Here we are testing that the translate service is called, but actually it is going above and beyond that and also testing that the UI is updated with the translated text.

In an ideal unit test world, we should have two unit tests, one that checks that the translate service is called and another that checks the UI shows the translated text.

I don't particularly care about the splitting out of the unit tests as there are pros and cons for that, but if we are combining tests like we are here, then the name of the test should be the most encompassing. Something like, "should show translated text on open accordion button". That way, the test for calling the translate service can be inferred from the test name.

omio-ma: - Initially was using TDD to force me to add that property
- atleast from what I have read from jasmine and spy, using line 46-49 would return an expected value when its called with the args passed there, unless I am wrong?

AWildBugAppeared: I'll give you a call as I think my explanation is lacking 😂

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/select-hearing-layout/select-hearing-layout.component.spec.ts
```diff
@@ -34,20 +37,44 @@ describe('SelectHearingLayoutComponent', () => {
...
     });
 
     it('should translate button text on text click', () => {
         component.ngOnInit();
         textButton.innerHTML = 'Close all';
```
FarronF, line 66: I don't think this line effects the test in any way does it?

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/select-hearing-layout/select-hearing-layout.component.spec.ts
```diff
@@ -34,20 +37,44 @@ describe('SelectHearingLayoutComponent', () => {
...
         textButton.click();
         expect(textButton.innerHTML).toBe('<span>select-hearing-layout.close-all</span>');
+        expect(component.currentButtonContentKey).toBe(buttonContentKeyWhenClosed);
```
FarronF, line 69: I think it would be better to use spyOn(component, 'setAccordionText') and expect that it is called correctly.

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/select-hearing-layout/select-hearing-layout.component.ts
```diff
@@ -25,9 +26,22 @@ export class SelectHearingLayoutComponent implements OnInit {
...
+
+        this.translateService.onLangChange.subscribe(event => {
+            const updatedHeadingElement = document.getElementById('accordion-choose-layout-heading');
+            const currentHeaderText = updatedHeadingElement.innerText;
+            const updatedHeaderText = this.translateService.instant('select-hearing-layout.choose-hearing-layout');
+
+            updatedHeadingElement.innerHTML = updatedHeadingElement.innerHTML.replace(currentHeaderText, updatedHeaderText);
+            const currentTextValue = this.accordionOpenAllElement.innerText.split('\n')[0];
+            const translatedElement = this.translateService.instant(`select-hearing-layout.${this.currentButtonContentKey}`);
+            this.accordionOpenAllElement.innerHTML = this.accordionOpenAllElement.innerHTML.replace(currentTextValue, translatedElement);
```
FarronF, line 43: Does this not already get set by calling setAccordianText earlier in the function?

FarronF, line 43: Can we call setAccordionText here?

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/select-hearing-layout/select-hearing-layout.component.spec.ts
```diff
@@ -34,20 +37,44 @@ describe('SelectHearingLayoutComponent', () => {
...
+    });
+
+    it('should show translated text on open/close toggle button', () => {
+        const expectedTranslatedContentForHeader = 'this is translated for the accordion header';
+        translateServiceSpy.instant
+            .withArgs('select-hearing-layout.choose-hearing-layout')
+            .and.returnValue(expectedTranslatedContentForHeader);
+        component.ngOnInit();
+        onLangChangeSpy.emit({ lang: 'tl' } as LangChangeEvent);
+        expect(headingButton.innerHTML).toContain(expectedTranslatedContentForHeader);
```
FarronF, line 61: Make it more clear that these are testing onLangChange event. I would nest these inside a describe.

`describe('onLangChange event', () => {
...
});`

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/select-hearing-layout/select-hearing-layout.component.ts
```diff
@@ -25,9 +26,22 @@ export class SelectHearingLayoutComponent implements OnInit {
...
+        });
     }
 
     setAccordionText(event: MouseEvent) {
```
FarronF, line 47: Remove MouseEvent since we don't use it.

omio-ma: which mouse event are you referring to? Do you mean line 47?

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/select-hearing-layout/select-hearing-layout.component.spec.ts
omio-ma: i will nest it in a describe block cool.

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/select-hearing-layout/select-hearing-layout.component.ts
omio-ma: the logic is a bit different as it changes the dom element itself so the same string split can't be used, initially I was trying to add the set accordion text funciton

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/select-hearing-layout/select-hearing-layout.component.spec.ts
omio-ma: should I make this change as it was an existing test, I added that assertion as I introduced that new property for this component? However it is tested in multiple places

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/select-hearing-layout/select-hearing-layout.component.spec.ts
omio-ma: you are right :) removed the emit function

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/select-hearing-layout/select-hearing-layout.component.ts
FarronF: Yes, I thought the comment was associated with that line.

omio-ma:
\azp run

omio-ma:
\azp run

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/select-hearing-layout/select-hearing-layout.component.ts
omio-ma: sorry missed this message, even is still being passed from line 28, and this isn't something I have added by the way

PR #1446, opened 2021-06-15
VIH-7828 date translated to welsh runtime

alberto-hortiguela:
https://tools.hmcts.net/jira/browse/VIH-7828

### Change description ###
DatePattern enum added
DatePipe for Translation Service added
DatePipe formatting pattern added
JOH, Judge, Participant components/html updated to translate date string run time

```
[ ] Yes
[ x ] No
```

### DIFF
--- a/VideoWeb/VideoWeb/ClientApp/src/app/services/models/date-patterns.ts
+++ b/VideoWeb/VideoWeb/ClientApp/src/app/services/models/date-patterns.ts
@@ -1,0 +1,4 @@
+export enum DatePatternOption {
+    Date = 'dd MMMM yyyy',
+    FullDate = 'fullDate'
+}
--- a/VideoWeb/VideoWeb/ClientApp/src/app/services/models/date-patterns.ts
+++ b/VideoWeb/VideoWeb/ClientApp/src/app/services/models/date-patterns.ts
@@ -1,4 +1,0 @@
-export enum DatePatternOption {
-    Date = 'dd MMMM yyyy',
-    FullDate = 'fullDate'
-}
--- a/VideoWeb/VideoWeb/ClientApp/src/app/shared/pipes/ngx-date.pipe.ts
+++ b/VideoWeb/VideoWeb/ClientApp/src/app/shared/pipes/ngx-date.pipe.ts
@@ -3,16 +3,14 @@ import { Pipe, PipeTransform } from '@angular/core';
 import { TranslateService } from '@ngx-translate/core';
 
 @Pipe({
-  name: 'DateLanguage'
+    name: 'translateDate',
+    pure: false
 })
 export class NgxDatePipe implements PipeTransform {
- 
-  constructor(private translateService: TranslateService) {
-  }
- 
-  public transform(value: any, pattern: string = 'fullDate'): any {
-    const datePipe = new DatePipe(this.translateService.currentLang);
-    console.log(datePipe);
-    return datePipe.transform(value, pattern);
-  }
+    constructor(private translateService: TranslateService) {}
+
+    public transform(value: any, pattern: string): any {
+        const datePipe = new DatePipe(this.translateService.currentLang);
+        return datePipe.transform(value, pattern);
+    }
 }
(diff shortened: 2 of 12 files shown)


VideoWeb/VideoWeb/ClientApp/src/app/services/models/date-patterns.ts
```diff
@@ -0,0 +1,4 @@
+export enum DatePatternOption {
+    Date = 'dd MMMM yyyy',
```
omio-ma, line 2: is there a complication if a class is used instead of creating this as an enum?

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/judge-waiting-room/judge-waiting-room.component.ts
```diff
@@ -22,13 +22,16 @@ import { NotificationToastrService } from '../services/notification-toastr.servi
...
 import { WaitingRoomBaseDirective } from '../waiting-room-shared/waiting-room-base.component';
+import { DatePatternOption } from 'src/app/services/models/date-patterns';
 
 @Component({
     selector: 'app-judge-waiting-room',
     templateUrl: './judge-waiting-room.component.html',
     styleUrls: ['./judge-waiting-room.component.scss', '../waiting-room-global-styles.scss']
 })
 export class JudgeWaitingRoomComponent extends WaitingRoomBaseDirective implements OnInit, OnDestroy {
+    public datePatternOption = DatePatternOption;
```
omio-ma, line 33: I am fairly new in angular, but shouldn't this be an interface and then the value could be set in the class? 

Or was it already considered and you would have repetition so instead enum was used?

AWildBugAppeared: The choice of enum is the right one as a DatePatternOption variable is one of x  date pattern options.

However, this line is not required. As I understand, this line is here so you can do 

`this.datePatternPipe = this.datePatternOption.Date;` 

on line 86, but you can just do

`this.datePatternPipe = DatePatternOption.Date;` 

If you could change that and the other files which use this pattern that would get rid of some unneeded lines. If I've missed something please let me know 👍

omio-ma: yea you can argue that we can set the value of props in an enum to be however we like, I am just used to seeing it with generic number values.

VideoWeb/VideoWeb/ClientApp/src/app/shared/pipes/ngx-date.pipe.ts
```diff
@@ -0,0 +1,16 @@
...
+import { TranslateService } from '@ngx-translate/core';
+
+@Pipe({
+    name: 'translateDate',
+    pure: false
+})
+export class NgxDatePipe implements PipeTransform {
+    constructor(private translateService: TranslateService) {}
+
+    public transform(value: any, pattern: string): any {
```
FarronF, line 12: I think pattern should be format to match DatePipe

https://angular.io/api/common/DatePipe

FarronF: As mentioned elsewhere I'm not sure if Enum is the way to go, looking into that now. But if we do can this be changed from a string to that Enum?

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/judge-waiting-room/judge-waiting-room.component.ts
alberto-hortiguela: I will remove the enum options as we just need to pass the string pattern on the Pipe...this was over complicating things itself

alberto-hortiguela:
Build pipeline issues
SonarCloud retries check potential observation/issue

PR #1447, opened 2021-06-16
VIH-7828 date translated to welsh runtime

alberto-hortiguela:
https://tools.hmcts.net/jira/browse/VIH-7828

### Change description ###
Change description
DatePipe for Translation Service added
DatePipe formatting pattern added
JOH, Judge, Participant components/html updated to translate date string run time

```
[ ] Yes
[ x ] No
```

### DIFF
--- a/VideoWeb/VideoWeb/ClientApp/src/app/shared/pipes/ngx-date.pipe.spec.ts
+++ b/VideoWeb/VideoWeb/ClientApp/src/app/shared/pipes/ngx-date.pipe.spec.ts
@@ -1,0 +1,21 @@
+import { NgxDatePipe } from './ngx-date.pipe';
+import { translateServiceSpy } from 'src/app/testing/mocks/mock-translation.service';
+
+describe('DatePipe', () => {
+    const translateService = translateServiceSpy;
+    const today = new Date('2021-06-16');
+
+    beforeEach(() => {
+        translateService.instant.calls.reset();
+    });
+
+    it('should return date format', () => {
+        const pipe = new NgxDatePipe(translateService);
+        expect(pipe.transform(today, 'dd MMM yyyy')).toBe('16 Jun 2021');
+    });
+
+    it('should return fullDate format', () => {
+        const pipe = new NgxDatePipe(translateService);
+        expect(pipe.transform(today, 'fullDate')).toBe('Wednesday, June 16, 2021');
+    });
+});
--- a/VideoWeb/VideoWeb/ClientApp/src/app/shared/pipes/ngx-date.pipe.spec.ts
+++ b/VideoWeb/VideoWeb/ClientApp/src/app/shared/pipes/ngx-date.pipe.spec.ts
@@ -1,16 +1,7 @@
 import { NgxDatePipe } from './ngx-date.pipe';
 import { translateServiceSpy } from 'src/app/testing/mocks/mock-translation.service';
 
-describe('DatePipe', () => {
-    const translateService = translateServiceSpy;
-    const today = new Date('2021-06-16');
-
-    beforeEach(() => {
-        translateService.instant.calls.reset();
-    });
-
-    it('should return date format', () => {
-describe('DatePipe', () => {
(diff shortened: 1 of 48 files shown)


VideoWeb/VideoWeb/ClientApp/src/app/shared/pipes/ngx-date.pipe.ts
```diff
@@ -0,0 +1,16 @@
...
+@Pipe({
+    name: 'translateDate',
+    pure: false
+})
+export class NgxDatePipe implements PipeTransform {
+    constructor(private translateService: TranslateService) {}
+
+    public transform(value: any, pattern: string): any {
+        const datePipe = new DatePipe(this.translateService.currentLang);
+        return datePipe.transform(value, pattern);
```
FarronF, line 14: This needs to be tested. It probably passed the gate before because there was enough lines of new code that were touched by other tests to pass. Also again very minor but can we change pattern to format same as DatePipe.

VideoWeb/VideoWeb/ClientApp/src/app/shared/pipes/ngx-date.pipe.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+    const translateService = translateServiceSpy;
+    const today = new Date('2021-06-16');
+
+    beforeEach(() => {
+        translateService.instant.calls.reset();
+    });
+
+    it('should return date format', () => {
+        const pipe = new NgxDatePipe(translateService);
+        expect(pipe.transform(today, 'dd MMM yyyy')).toBe('16 Jun 2021');
```
FarronF, line 14: This is really testing DatePipe. For our test we should just ensure we pass and return the correct values and not actually hit DatePipe. I can show you how to do this.

FarronF: But actually looking at it, it's quite hard to properly unit test this pipe properly as we create a new DatePipe each call. Strictly speaking we should inject a factory to create that.

FarronF: This is a more isolated version but still not perfect as we can't ensure the correct language is passed.

```suggestion
describe('DatePipe', () => {
    const translateService = translateServiceSpy;
    const testDate = new Date('2021-06-16');
    const testFormat = 'TestFormat';

    let pipe: NgxDatePipe;
    const expectedValue = 'expectedValue';

    beforeEach(() => {
        translateService.instant.calls.reset();
        pipe = new NgxDatePipe(translateService);
        spyOn(DatePipe.prototype, 'transform').and.returnValue('expectedValue');
    });

    it('should return correct value from DatePipe', () => {
        const value = pipe.transform(testDate, testFormat);
        expect(DatePipe.prototype.transform).toHaveBeenCalledWith(testDate, testFormat);
        expect(DatePipe.prototype.transform).toHaveBeenCalledTimes(1);
        expect(value).toBe(expectedValue);
    });
});
```

VideoWeb/VideoWeb/ClientApp/src/app/shared/pipes/ngx-date.pipe.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+    const testDate = new Date('2021-06-16');
+    const testFormat = 'TestFormat';
+
+    let pipe: NgxDatePipe;
+    const expectedValue = 'expectedValue';
+
+    beforeEach(() => {
+        translateService.instant.calls.reset();
+        pipe = new NgxDatePipe(translateService);
+        spyOn(NgxDatePipe.prototype, 'transform').and.returnValue('expectedValue');
```
FarronF, line 15: This should be DatePipe. We use spyOn to prevent it from calling code external to what we want to test. This is overriding the function we want to test which is why the coverage is failing.

robertbennett1998 (approved):
LGTM

PR #1464, opened 2021-07-02
VIH-7917 iPad users on Safari are seeing the "unsupported device" page

ZJianWang:

VIH-7917

### Change description ###

we are looking for "mac" os rather than "ios" in isIpad method, but I think we still need "mac" for some other apple devices so I added "ios"

```
[ ] Yes
[x] No
```


alberto-hortiguela (approved):
LGTM

PR #1480, opened 2021-07-19
VIH-7988 persist current rooms and vmrs

FarronF:
VIH-7988

### Change description ###
Get the vmr from the cache. Do not override the current room as we can't get this from the cache currently. Set interpreter ID based on existing participant id if can be found.

```
[ ] Yes
[x] No
```


VideoWeb/VideoWeb/Mappings/ParticipantToParticipantResponseMapper.cs
```diff
@@ -9,31 +9,33 @@
...
-                Id = input.Id,
-                InterpreterRoom = null,
-                LastName = input.LastName,
-                Name = input.Name,
-                Representee = input.Representee,
-                Role = input.Role,
-                Status = input.ParticipantStatus,
-                LinkedParticipants = input.LinkedParticipants.Select(x => linkedParticipantMapper.Map(x)).ToList(),
+                CaseTypeGroup = participant.CaseTypeGroup,
+                CurrentRoom = null, // This cannot currently be gotten from the conference cache, the UI will keep the current room for an existing user.
```
robertbennett1998, line 27: Have you raised a tech debt ticket for this?

VideoWeb/VideoWeb.UnitTests/Controllers/InternalEventController/ParticipantsUpdatedTests.cs
```diff
@@ -77,8 +77,9 @@ public void Setup()
                 .Setup(x => x.Get(It.Is<EventType>(eventType => eventType == EventType.ParticipantsUpdated)))
                 .Returns(_mocker.Mock<IEventHandler>().Object);
 
-            _mocker.Mock<IMapperFactory>().Setup(x => x.Get<ParticipantRequest, Participant>()).Returns(_mocker.Create<ParticipantRequestMapper>());
-            _mocker.Mock<IMapperFactory>().Setup(x => x.Get<Participant, ParticipantResponse>()).Returns(_mocker.Create<ParticipantToParticipantResponseMapper>());
+            _mocker.Mock<IMapperFactory>().Setup(x => x.Get<ParticipantRequest, IEnumerable<Participant>, Participant>()).Returns(_mocker.Create<ParticipantRequestMapper>());
```
robertbennett1998, line 80: I don't think these are necessary. I think auto mock should create these for you anyways.

robertbennett1998 (approved):
LGTM. Just double check the two comments. Only minor and do not effect merging.

VideoWeb/VideoWeb/Mappings/ParticipantToParticipantResponseMapper.cs
FarronF: We could either add it to the cache, or change this so that the message comes from Video Api when everything is finished. The second option was discussed but has not been started. I will see if there was an actual ticket created for that.

PR #1483, opened 2021-07-20
VIH-8000 add footer translations to unsupported browser page

alberto-hortiguela:
https://tools.hmcts.net/jira/browse/VIH-8000

### Change description ###
Update welsh content on unsupported web page
Footer link to be displayed always

```
[ ] Yes
[x] No
```


VideoWeb/VideoWeb/ClientApp/src/app/shared/footer/footer.component.ts
```diff
@@ -14,12 +14,10 @@ export class FooterComponent implements OnInit {
     hideContactUsLink = false;
     privacyPolicyUri = pageUrls.PrivacyPolicy;
     accessibilityUri = pageUrls.Accessibility;
-    hideLinksForUnsupportedBrowser = false;
 
     constructor(private router: Router, private translate: TranslateService, private logger: Logger) {
         this.router.events.pipe(filter((event: RouterEvent) => event instanceof NavigationEnd)).subscribe(x => {
```
FarronF, line 19: Not a serious issue since the footer will always be there but we should probably still unsubscribe onDestroy.

alberto-hortiguela: Added on last commit

VideoWeb/VideoWeb/ClientApp/src/app/shared/footer/footer.component.ts
```diff
@@ -1,38 +1,39 @@
...
     }
 
     ngOnInit() {
         this.hideContactUs();
     }
 
-    hideContactUs() {
-        this.hideContactUsLink = this.router.url === '/contact-us';
+    ngOnDestroy(): void {
+        this.routerEventsSubscription$.unsubscribe();
```
FarronF, line 32: This will work but we should probably settle on a pattern. I like the takeUntil onDestroy pattern @robertbennett1998  has used in some of the changes on the waiting room components. We can also look at other solutions such as this https://github.com/ngneat/until-destroy

JohWaitingRoomComponent
JudgeWaitingRoomComponent
ParticipantWaitingRoomComponent

VideoWeb/VideoWeb/ClientApp/src/app/shared/footer/footer.component.ts
```diff
@@ -1,38 +1,39 @@
...
     }
 
-    hideLinks() {
-        this.hideLinksForUnsupportedBrowser = this.router.url === `/${pageUrls.UnsupportedBrowser}`;
```
FarronF, line 35: Was this not the point of this story? Or is it done a different way now?

FarronF: I misunderstood this. I thought you had added this in an earlier commit.

FarronF (approved):
This seems fine. Just added a comment about how we unsubscribe but that's a bigger discussion to have.

VideoWeb/VideoWeb/ClientApp/src/app/shared/footer/footer.component.ts
robertbennett1998: I agree we need to make this consistent and I think the takeUnitl approach or the package should be used as it is more explicit.

PR #1485, opened 2021-07-21
VIH-8003 consulation judicial participants

FarronF:
VIH-8003

### Change description ###
Do not update the status as part of the ParticipantsUpdated message as the cache is not a reliable source.

```
[ ] Yes
[x] No
```

### DIFF
--- a/VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/waiting-room-shared/waiting-room-base.component.ts
+++ b/VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/waiting-room-shared/waiting-room-base.component.ts
@@ -157,7 +157,6 @@ export abstract class WaitingRoomBaseDirective {
     }
 
     getLoggedParticipant(): ParticipantResponse {
-        console.log('Faz - this.loggedInUser', this.loggedInUser);
         return this.conference.participants.find(x => x.id === this.loggedInUser.participant_id);
     }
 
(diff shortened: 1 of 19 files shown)


VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/waiting-room-shared/waiting-room-base.component.ts
```diff
@@ -458,25 +458,27 @@ export abstract class WaitingRoomBaseDirective {
...
                         this.participant.status === ParticipantStatus.InHearing
                     );
                 });
 
                 const updatedParticipants = [...participantsUpdatedMessage.participants].map(updatedParticipant => {
-                    updatedParticipant.current_room = this.conference.participants.find(
-                        currentParticipant => currentParticipant.id === updatedParticipant.id
-                    )?.current_room;
+                    const currentParticipant = this.conference.participants.find(x => x.id === updatedParticipant.id);
+                    updatedParticipant.current_room = currentParticipant ? currentParticipant.current_room : null;
```
ZJianWang, line 475: if the current_room is null means couldn't get the room.label, I'm considering this will cause some css issue but I might wrong

FarronF: This is only for new participants who will not have a room at the start. Currently they will have an undefined room anyway but null evaluates a little faster(this is negligible but still there).

I actually would like to change the general structure of participants, as this model has everything, it would be nice for it to be split between details which don't change except via admin web and state such as room and status which change constantly via pexip/our own events.

ZJianWang (approved):
LGTM

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/waiting-room-shared/waiting-room-base.component.ts
FarronF: The updatedParticipant name might be a bit confusing here. I've changed it in a later commit.

robertbennett1998 (approved):
LGTM

PR #1490, opened 2021-07-27
VIH-7852 skip link accessability

FarronF:
VIH-7852

### Change description ###
Moving the waiting room link and the back links in self test into the header that is controlled by a service.

```
[ ] Yes
[x] No
```


FarronF:
Unit tests need to be updated. Not totally happy with all the super calls that are required with this implementation either. Will look into other possibilities.

FarronF:
This will most likely not be going in. I prefer the implementation in https://github.com/hmcts/vh-video-web/pull/1493. I will be closing this if I do go with that one

PR #1511, opened 2021-08-11
VIH-8083 Fixed console errors on judge navigation back to hearing list

robertbennett1998:
https://tools.hmcts.net/jira/browse/VIH-8083

### Change description ###
- CORS errors not fixed.
- Fixed errors from participant, conference and video control cache services where the conference was undefined due to navigating away from a conference specific page.

```
[ ] Yes
[x] No
```

### DIFF
--- a/VideoWeb/VideoWeb/ClientApp/src/app/services/conference/participant.service.ts
+++ b/VideoWeb/VideoWeb/ClientApp/src/app/services/conference/participant.service.ts
@@ -105,7 +105,6 @@ export class ParticipantService {
         private videoControlCacheService: VideoControlCacheService,
         private logger: LoggerService
     ) {
-        this.logger.warn(`${this.loggerPrefix} Constructor called.`);
         this.initialise();
     }
 
@@ -149,6 +148,11 @@ export class ParticipantService {
             this.logger.info(`${this.loggerPrefix} new conference`, {
                 conference: conference
             });
+
+            if (!conference) {
+                this.logger.warn(`${this.loggerPrefix} no conference loaded; skipping initialisation.`);
+                return;
+            }
 
             this._endpointParticipants = [];
             this._nonEndpointParticipants = [];
--- a/VideoWeb/VideoWeb/ClientApp/src/app/services/conference/participant.service.ts
+++ b/VideoWeb/VideoWeb/ClientApp/src/app/services/conference/participant.service.ts
@@ -149,14 +149,14 @@ export class ParticipantService {
                 conference: conference
             });
 
+            this._endpointParticipants = [];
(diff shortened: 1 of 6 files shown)


VideoWeb/VideoWeb/ClientApp/src/app/services/conference/video-control-cache.service.ts
```diff
@@ -18,6 +18,11 @@ export class VideoControlCacheService {
...
     ) {
         this.conferenceService.currentConference$.subscribe(conference => {
+            if (!conference) {
+                this.logger.warn(`${this.loggerPrefix} No conference loaded. Skipping loading of hearing state for conference`);
+                return;
+            }
+
             this.storageService.loadHearingStateForConference(conference.id).subscribe(state => {
                 this.hearingControlStates = state;
             });
```
FarronF, line 28: I know this is existing code. But does this not create a new subscription every time the currentConference changes that never completes?

VideoWeb/VideoWeb/ClientApp/src/app/services/conference/video-control-cache.service.ts
```diff
@@ -18,6 +18,11 @@ export class VideoControlCacheService {
         private logger: LoggerService
     ) {
         this.conferenceService.currentConference$.subscribe(conference => {
+            if (!conference) {
+                this.logger.warn(`${this.loggerPrefix} No conference loaded. Skipping loading of hearing state for conference`);
+                return;
+            }
```
FarronF, line 24: Could you add a test when conference is falsy that checks that it doesn't go past this point?

VideoWeb/VideoWeb/ClientApp/src/app/services/conference/participant.service.ts
```diff
@@ -154,6 +153,11 @@ export class ParticipantService {
             this._nonEndpointParticipants = [];
             this._virtualMeetingRooms = [];
 
+            if (!conference) {
+                this.logger.warn(`${this.loggerPrefix} no conference loaded; skipping initialisation.`);
+                return;
+            }
+
```
FarronF, line 160: Same as in the other service it would be good to test it doesn't proceed when falsy

VideoWeb/VideoWeb/ClientApp/src/app/services/conference/video-control-cache.service.ts
robertbennett1998: Well spotted. Yeah, overlooked as it should be a call to an endpoint but currently it is using the cache. Will add a pipe.

kevin-hmcts (approved):
LGTM

PR #1513, opened 2021-08-12
VIH-7884 Add additional context to the display name to specify heartbeats

robertbennett1998:
https://tools.hmcts.net/jira/browse/VIH-7884

### Change description ###
- Updated client side model.
- Updated BFFE to add "HEARTBEAT" or "NO_HEARTBEAT" it based on role.
- Updated relevant unit tests.

```
[x] Yes
[ ] No
```


VideoWeb/VideoWeb/Helpers/ParticipantTilePositionHelper.cs
```diff
@@ -18,9 +18,11 @@ public static void AssignTilePositions(IEnumerable<ParticipantResponse> particip
         public static string GetTiledDisplayName(ParticipantResponse participant)
         {
             var prefix = "";
+            var heartbeatMode = "NO_HEARTBEAT";
```
omio-ma, line 21: declare the NO_HEARTBEAT and HEARTBEAT as consts either in this file, or in a separate file so you could use that to reference in your unit test

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/participants-panel/participants-panel.component.spec.ts
```diff
@@ -486,7 +486,18 @@ describe('ParticipantsPanelComponent', () => {
             const panelModel = component.participants[1];
             panelModel.updateParticipant(false, false, false);
 
-            const participantModel = new ParticipantModel('', '', '', '', null, null, null, false, null, null);
+            const participantModel = new ParticipantModel(
+                '',
+                '',
+                '',
+                'ROLE;NO HEARTBEAT;DISPLAY NAME;ID',
```
omio-ma, line 493: reference string const here

VideoWeb/VideoWeb/ClientApp/src/app/services/conference/models/pexip-display-name-model.spec.ts
```diff
@@ -0,0 +1,125 @@
+import { invalidPexipDisplayNameFormatError } from '../errors/invalid-pexip-display-name-format.error';
+import { HeartbeatMode } from './heartbeat-mode.model';
+import { PexipDisplayNameModel } from './pexip-display-name.model';
+
+describe('PexipDisplayNameModel', () => {
+    beforeEach(() => {
+        return;
```
omio-ma, line 7: is this left by mistake?

VideoWeb/VideoWeb.UnitTests/Helpers/ParticipantTilePositionHelperTest.cs
```diff
@@ -42,8 +28,22 @@ private List<ParticipantResponse> GetParticipantResponses()
...
                 new ParticipantResponse { Id = Guid.NewGuid(), Role = Role.Individual, DisplayName = "Part7", CaseTypeGroup = "group1", HearingRole = "Applicant" }
             };
+        }
+
+        [Test]
+        public void Should_set_tailed_display_names_for_non_judge_participants_number_participant()
+        {
+            ParticipantTilePositionHelper.AssignTilePositions(_participants);
+
+            _participants[0].TiledDisplayName.Should().Be($"JUDGE;HEARTBEAT;{_participants[0].DisplayName};{_participants[0].Id}");
```
omio-ma, line 38: We should either assert if the display name property contains the participant display name and id, or we can store this magic string in a separate const and leave the assertions as is

robertbennett1998: The string has to be in the correct order so doing a contain isn't specific enough. What do you mean by "can store this magic string in a separate const and leave the assertions as is"?

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/participants-panel/participants-panel.component.spec.ts
robertbennett1998: Changed to null it wasn't needed

VideoWeb/VideoWeb.UnitTests/Helpers/ParticipantTilePositionHelperTest.cs
omio-ma: It looks like the first part of this string is related to the role that we get from pariticpantResponse, can we reference the role here instead? 
For the heartbeat string we can probably declare two seperate consts and reference it here, so basically it will be like this:
$"{_participants[0].Role.toUppercase()};{heartbeat};{_participants[0].DisplayName};{_participants[0].Id}"

But do we need to unit test the complete TiledDisplayName here? Can we just check for the variables being displayed here instead, so using the Contain to check for the properties? 

_participants[0].TiledDisplayName.Should().Contain(_participants[0].DisplayName);

robertbennett1998: The role doesn't directly relate to that Role enum i.e. Individual -> Civilian; JOH -> Civilian. I have raised a refactor ticket about the ParticipantTilePositionHelper as well.

Yeah as this class has to be correct as it is what populates it for the front end. If the ID and Display name were the wrong way round. None of Kinly's events would work and the display name would be the participants ID. Using contains doesn't check the order of the role, heartbeat mode, display name and ID.

ZJianWang (approved):
LGTM

omio-ma (approved):

PR #1514, opened 2021-08-17
VIH-8114 - fix tests affected by UI change

MichaelKeaneathmcts:
### Change description ###

```
[ ] Yes
[ ] No
```


VideoWeb/VideoWeb.AcceptanceTests/Pages/HearingRoomPage.cs
```diff
@@ -15,6 +15,6 @@ public static class HearingRoomPage
         public static readonly By ToggleHandRaised = By.Id("toggle-hand-raised-img");
         public static readonly By HandRaised = By.ClassName($"fa-hand-paper");
         public static readonly By ParticipantPanel = By.Id($"panelList");
-        public static By InterPreterName(string displayName) => By.XPath($"//span[contains(text(),'{displayName}')]");
+        public static By InterPreterName(string displayName) => By.XPath($"//span[contains(text()[2], '{displayName}')]");
```
FarronF, line 18: Would it be worth adding ids to these elements on the UI to make these tests easier?

VideoWeb/VideoWeb.AcceptanceTests/Hooks/ConfigHooks.cs
```diff
@@ -28,7 +28,7 @@ public class ConfigHooks
 
         public ConfigHooks(TestContext context)
         {
-            _configRoot = ConfigurationManager.BuildConfig("CA353381-2F0D-47D7-A97B-79A30AFF8B86", "18c466fd-9265-425f-964e-5989181743a7");
+             _configRoot = ConfigurationManager.BuildConfig("CA353381-2F0D-47D7-A97B-79A30AFF8B86", "18c466fd-9265-425f-964e-5989181743a7");
```
FarronF, line 31: This whitespace shouldn't be there.

FarronF (approved):

VideoWeb/VideoWeb.AcceptanceTests/Pages/HearingRoomPage.cs
MichaelKeaneathmcts: It would be, I asked Simon and kevin if this could be done for all web elements, I'd also want there to be a rule that you don't put two values in one web element as this does: <span>Auto_individual_44<br>Auto_interpreter_04<span> as that makes finding the elements harder and the testing becomes brittle as it could change at any time,

alberto-hortiguela (approved):
LGTM

VideoWeb/VideoWeb.AcceptanceTests/Pages/HearingRoomPage.cs
PrasannaDommalapati: if using xpath without using hardcoded index is ok.

PR #1529, opened 2021-08-26
VIH-8041  judge blur background

shaed-parkar:

VIH-8041 
VIH-8044

### Change description ###

* Allow judges to blur their background or select an image as their background
  * During hearing and consultation
  * Before entering a hearing

```
[ ] Yes
[x] No
```


VideoWeb/VideoWeb/ClientApp/.browserslistrc
```diff
@@ -6,4 +6,4 @@
 last 2 versions
 Firefox ESR
 not dead
-IE 9-11
\ No newline at end of file
+# IE 9-11
```
FarronF, line 9: I believe this was uncommented to show the unsupported browser page on IE.

VideoWeb/VideoWeb/ClientApp/src/app/app-routing.module.ts
```diff
@@ -29,6 +29,6 @@ export const routes: Routes = [
 
 @NgModule({
     exports: [RouterModule],
-    imports: [RouterModule.forRoot(routes, { enableTracing: !environment.production })]
+    imports: [RouterModule.forRoot(routes, { enableTracing: environment.production })]
```
FarronF, line 32: This should only be for debugging. 

Is the angular environment always set to production?

VideoWeb/VideoWeb/ClientApp/src/app/services/user-media-stream.service.spec.ts
```diff
@@ -9,29 +9,44 @@ import { mustProvideAMicrophoneDeviceError, UserMediaStreamService } from './use
...
+        stream.clone.and.callFake(() => {
+            cloneStream = Object.assign({}, stream);
+            return cloneStream;
+        });
         return stream;
     };
 
     const cameraOneDevice = new UserMediaDevice('Camera 1', Guid.create().toString(), 'videoinput', '');
-    const cameraOneStream = mediaStreamBuilder(cameraOneDevice);
+    let cameraOneStream = mediaStreamBuilder(cameraOneDevice);
```
FarronF, line 40: You could just declare these here since they are assigned to in the beforeEach.

VideoWeb/VideoWeb/ClientApp/.browserslistrc
shaed-parkar: We can still leave this line as is. The config is there to ensure the build works for older browsers. If we're only showing an unsupported page I don't think we need the whole app to be checked against IE.

I would be happy to revert this though

VideoWeb/VideoWeb/ClientApp/src/app/app-routing.module.ts
shaed-parkar: Thanks! i forgot to undo this.

VideoWeb/VideoWeb/ClientApp/src/app/services/user-media-stream.service.spec.ts
shaed-parkar: moved it to the beforeEach

VideoWeb/VideoWeb/ClientApp/src/app/services/video-filter.service.ts
```diff
@@ -0,0 +1,172 @@
...
+    activeFilter: BackgroundFilter;
+    imgs: Map<BackgroundFilter, HTMLImageElement> = new Map();
+
+    constructor(private logger: Logger) {
+        this.filterOn = false;
+        this.activeFilter = null;
+
+        this.selfieSegmentation = new SelfieSegmentation({
+            locateFile: file => {
+                return `https://cdn.jsdelivr.net/npm/@mediapipe/selfie_segmentation@0.1.1628007100/${file}`;
```
FarronF, line 36: Any way we could change this to something we have control over?

shaed-parkar: What do you mean? Specifying the version number is total control. Also, this number should change in line with the version in the package.json file

VideoWeb/VideoWeb/ClientApp/src/app/services/video-filter.service.ts
```diff
@@ -0,0 +1,185 @@
...
+        if (this.canvasStream) {
+            return;
+        }
+        this.logger.debug(`${this.loggerPrefix} initialising stream for filter`);
+        this.videoElement = document.createElement('video');
+        this.videoElement.srcObject = stream;
+
+        this.canvasElement = document.createElement('canvas');
+        this.canvasElement.width = 1280;
+        this.canvasElement.height = 720;
```
FarronF, line 71: Are these always the same? Could they change if someone is using phone? 

I see we also have the same width and height on lines 86/87. Do these need to be the same values? If so could we have them as consts?

VideoWeb/VideoWeb/ClientApp/src/app/services/video-filter.service.ts
```diff
@@ -0,0 +1,185 @@
...
+                break;
+        }
+    }
+
+    private applyBlurEffect(results: Results, withBlur: boolean) {
+        if (withBlur) {
+            this.canvasCtx.filter = 'blur(50px)';
+        }
+        this.canvasCtx.drawImage(results.image, 0, 0, this.canvasElement.width, this.canvasElement.height);
+    }
```
FarronF, line 156: Maybe the name of the method could be clearer, feel it implies it will always have a blur.

VideoWeb/VideoWeb/ClientApp/src/app/shared/video-filter/video-filter.component.html
```diff
@@ -0,0 +1,85 @@
+<button class="vh-accordion" (click)="toggleDisplayOptions()">
+  <span class="govuk-link">
+    {{ 'video-filter.title' | translate }}
+    <i
+      class="fas"
+      [ngClass]="showOptions ? 'fa-chevron-up' : 'fa-chevron-down'"
+      id="bg-icon"
+      style="margin-left: 8px"
```
FarronF, line 8: Put style in css.

VideoWeb/VideoWeb/ClientApp/src/app/shared/video-filter/video-filter.component.scss
```diff
@@ -0,0 +1,96 @@
...
+  position: absolute;
+  left: -9999px;
+}
+
+input[type='radio']:checked + label > img {
+  border: 4px solid govuk-colour('yellow');
+}
+
+//Grid for background images
+.bg-grid {
```
FarronF, line 50: I'd prefer if bg was background. 

How does this grid look on smaller screens? I'd also wonder about if other options are added, they will have to be added here as well, might be a way other than grid to do.

VideoWeb/VideoWeb/ClientApp/src/app/shared/video-filter/video-filter.component.ts
```diff
@@ -0,0 +1,55 @@
...
+
+@Component({
+    selector: 'app-video-filter',
+    templateUrl: './video-filter.component.html',
+    styleUrls: ['./video-filter.component.scss']
+})
+export class VideoFilterComponent implements OnInit, OnDestroy {
+    destroy$: Subject<boolean> = new Subject<boolean>();
+    private readonly loggerPrefix = '[VideoFilter] -';
+    vBG = BackgroundFilter;
```
FarronF, line 16: More descriptive name.

VideoWeb/VideoWeb/ClientApp/src/app/shared/select-media-devices/select-media-devices.component.scss
```diff
@@ -45,12 +45,6 @@
 .wr-h {
   margin-left: 25px;
 }
-/*mirrored video*/
-#preferredCameraStream {
-  transform: rotateY(180deg);
-  -webkit-transform: rotateY(180deg); /* Safari and Chrome */
-  -moz-transform: rotateY(180deg); /* Firefox */
-}
```
FarronF, line 53: Does this still get mirrored another way?

FarronF (commented):
I think this looks really good. A few questions to clarify and some minor comments. But I would like to see it in action before approving so lets try to get a demo arranged.

VideoWeb/VideoWeb/ClientApp/src/app/services/video-filter.service.ts
shaed-parkar: What would you suggest? The usage of the method is explicitly only when the active filter is blur

VideoWeb/VideoWeb/ClientApp/src/app/shared/video-filter/video-filter.component.scss
shaed-parkar: It should be fine, I used the styles provided also develop with the minimum VH size requirements of 1366 x 768

VideoWeb/VideoWeb/ClientApp/src/app/shared/video-filter/video-filter.component.html
shaed-parkar: will do

VideoWeb/VideoWeb/ClientApp/src/app/shared/video-filter/video-filter.component.ts
shaed-parkar: sure

VideoWeb/VideoWeb/ClientApp/src/app/shared/select-media-devices/select-media-devices.component.scss
shaed-parkar: yes, via the video filter

VideoWeb/VideoWeb/ClientApp/.browserslistrc
FarronF: We can display something to IE users without having the app actually work for them just to display an unsupported browser page all right. I'm not entirely clear why the change was made in the first place, I believe it was a local issue so I'm ok with this as is and I'll bring it up again else where.

FarronF (approved):
Happy to approve after demo and changes made/questions clarified.

omio-ma (commented):
we need to scale down the size of these two images: VideoWeb/VideoWeb/ClientApp/src/assets/images/VhBgFilterHMCTS.jpg and VideoWeb/VideoWeb/ClientApp/src/assets/images/VhBgFilterSCTS.jpg, they are over 1 mb

shaed-parkar:
> we need to scale down the size of these two images: VideoWeb/VideoWeb/ClientApp/src/assets/images/VhBgFilterHMCTS.jpg and VideoWeb/VideoWeb/ClientApp/src/assets/images/VhBgFilterSCTS.jpg, they are over 1 mb

Resized the images to 1280x720 (sizes are lower than 200KB now)

FarronF (dismissed):

FarronF (approved):

omio-ma (approved):

PR #1543, opened 2021-09-06
VIH-8012 Toastr notification for quick links users.

FarronF:
VIH-8012

### Change description ###
Pushing event for UI when a quick link user is added to a hearing.

```
[ ] Yes
[x] No
```


VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/services/notification-toastr.service.ts
```diff
@@ -294,19 +294,24 @@ export class NotificationToastrService {
...
             }
         )}</span>`;
-        message += `<span class="toast-content toast-body">${this.translateService.instant(
-            'notification-toastr.participant-added.message',
-            {
-                role: this.translateHearingRole(participant.hearing_role),
-                party: this.translateCaseRole(participant.case_type_group)
-            }
-        )}</span>`;
+        message += `<span class="toast-content toast-body">${messageBody}</span>`;
```
omio-ma, line 314: does this not have a template file, can we move the html bit into it if we do?

VideoWeb/VideoWeb/Helpers/ParticipantsUpdatedEventNotifier.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        private Task PublishEventToUi(CallbackEvent callbackEvent)
+        {
+            if (callbackEvent == null)
+            {
+                return Task.CompletedTask;
+            }
+
+            var handler = _eventHandlerFactory.Get(callbackEvent.EventType);
+            return handler.HandleAsync(callbackEvent);
```
omio-ma, line 52: is this async just by name or actually an async method, if its the latter we should make this as async method and use await here

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/services/notification-toastr.service.ts
FarronF: Unfortunately no. This would be part of the toastr tech debt story if that ends up being done.

VideoWeb/VideoWeb/Helpers/ParticipantsUpdatedEventNotifier.cs
FarronF: Honestly I'm not totally clear on why it is this way. This came from internal events controller as I didn't want to duplicate this code in QuickLinks controller. And this is the same as in VideoEvents controller so maybe that should be put somewhere central itself. But this is how all the events are currently handled. Perhaps it is something that could be looked at but it wouldn't just be for this event.

FarronF: I'm not 100% sure on this but I would have thought this is ok because we are returning a task and it is awaited elsewhere by the code. In this case on line 41. Although I think this await should be in the controller rather than in the this class.

VideoWeb/VideoWeb.UnitTests/Helpers/ParticipantsUpdatedEventNotifierTests.cs
```diff
@@ -0,0 +1,85 @@
...
+
+            _conference = new Conference();
+            _conference.Id = Guid.NewGuid();
+            _conference.Participants = new List<Participant>();
+
+            _participant1 = new Participant();
+            _participant1.Id = Guid.NewGuid();
+
+            _participant2 = new Participant();
+            _participant2.Id = Guid.NewGuid();
```
krishnapottekkat-hmcts, line 46: Is this should be _participant2.Id ?

omio-ma (approved):

krishnapottekkat-hmcts (approved):
LGTM

PR #1546, opened 2021-09-07
VIH-8170 self view dimensions are not responsive in consultation/hearing room

robertbennett1998:
https://tools.hmcts.net/jira/browse/VIH-8170
Screenshots of Samsung Galaxy S20 and an iPad Air (2013) on the ticket.

### Change description ###
- Moved self view to its own component
- Set a min and max width for the self-view that is device independent

```
[ ] Yes
[x] No
```


PrasannaDommalapati (approved):
LGTM

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/private-consultation-room-controls/self-view/self-view.component.scss
```diff
@@ -0,0 +1,13 @@
...
+  border: 3px solid govuk-colour('yellow');
+}
+
+.video-container {
+  bottom: 10px;
+  left: 10px;
+  z-index: 101;
+  position: absolute;
+  max-width: 10vw;
+  min-width: 8rem;
```
FarronF, line 11: Just want to confirm the behavior here. I'm not entirely sure how this will behave due to the video element having a width of 100% but I would think the video shouldn't ever effect the width because of this. 

I believe min-width takes priority here does it ever get larger than 8rem? It would be good to get a test with cameras that have different aspect ratios.

kevin-hmcts (approved):
LGTM

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/private-consultation-room-controls/self-view/self-view.component.scss
robertbennett1998: "I believe min-width takes priority here does it ever get larger than 8rem? It would be good to get a test with cameras that have different aspect ratios."

It does increase but only on fairly large screens like 720p+ and very slight

PR #1551, opened 2021-09-08
VIH-8041 video filter improvements

shaed-parkar:

VIH-8041

### Change description ###

* use original video stream size instead of forcing 1280x720
* upgrading mediapipe selfie segmentation package to the latest version
* determine the ability to use filters with the new EnableVideoFilters config setting
* disable filtering options on tablets or Safari browsers

```
[ ] Yes
[x] No
```


FarronF (dismissed):

VideoWeb/VideoWeb/ClientApp/src/app/services/video-filter.service.ts
```diff
@@ -34,7 +38,8 @@ export class VideoFilterService {
     activeFilter: BackgroundFilter;
     imgs: Map<BackgroundFilter, HTMLImageElement> = new Map();
 
-    constructor(private logger: Logger) {
+    constructor(private logger: Logger, private configService: ConfigService, private deviceTypeService: DeviceTypeService) {
+        this.configService.getClientSettings().subscribe(settings => (this.enableVideoFilters = settings.enable_video_filters));
```
FarronF, line 42: If this never changes during use of the app could you put a pipe(first()) before the subscribe.

shaed-parkar: We could, but that's somewhat taken care of by the service itself

FarronF: Fair enough it's not needed for http calls since they complete themselves but I'm not clear how the config service works with the caching. I think it would be good to be explicit here.

FarronF (approved):
I'm not sure if what I commented on would actually be an issue but it would be negligible anyway and is generally how we did it so will approve.

omio-ma (approved):

PR #1554, opened 2021-09-08
VIH-8182 changed style for the full name input field

omio-ma:
### JIRA link ###
https://tools.hmcts.net/jira/browse/VIH-8182

### Change description ###
- fixed input field taking full width

```
[ ] Yes
[x] No
```


omio-ma:
I have decided against using the class provided in the ticket as its using `width:50%`, meaning that would change based on the devices, for example in ipad the input field will appear to be smaller than 1/2 the size. I made use of min width to avoid that and used media query to ensure it only appears for larger screen sizes, however if this is not deemed good practice I can change back to use the class provided.

omio-ma:
I have just confirmed the style with Neil and he is happy with this change over using the class provided

VideoWeb/VideoWeb/ClientApp/src/app/on-the-day/quick-links/quick-links.component.html
```diff
@@ -9,7 +9,7 @@ <h1 class="govuk-heading-xl govuk-!-margin-top-8">{{ 'quick-participant-landing-
       <label class="govuk-label" for="full-name">
         {{ 'quick-participant-landing-page.name-label' | translate }}
       </label>
-      <input formControlName="name" class="govuk-input" id="full-name" name="full-name" type="text" spellcheck="false" autocomplete="name">
+      <input formControlName="name" class="govuk-input full-name-text-input" id="full-name" name="full-name" type="text" spellcheck="false" autocomplete="name">
```
FarronF, line 12: Is there not an existing class in gov uk that we can use for this?

omio-ma: thats what I was discussing with Neil, it doesnt appear to be one, the one provided in the ticket wouldnt look good in ipad

FarronF: .govuk-grid-column-one-half

omio-ma: The class I was attempting to use was this one: govuk-input govuk-!-width-one-half, and it doesnt look good in ipad, the other class Neil wanted me to try out with didnt have media query so was using a fixed width for all media devices

FarronF: I didn't see your comment regarding why. This class should already have responsiveness for smaller devices.

VideoWeb/VideoWeb/ClientApp/src/app/on-the-day/quick-links/quick-links.component.scss
```diff
@@ -0,0 +1,6 @@
+@media (min-width: 40.0625em) {
+  .full-name-text-input {
+    width: 50%;
+    min-width: 25em;
```
FarronF, line 4: If we do stick with this solution please change this to rem.

FarronF: And just to be clear on devices from 641-800ish pixels this will be more than half width. Not accounting for padding or margins of parent elements. That's fine but just want it to be clear that Neil is ok with that.

omio-ma: when using rem, it appears to have the same effect as using that class that Neil didnt like as well, so shall I stick with rem or keep using the class which makes it look weird in ipad view?

FarronF: em isn't consistant enough for this sort of styling as it could be changed by a change to any of it's ancestor elements. We can change the number for rem if that is a concern.

Also perhaps this should be on the container div rather than the input itself and allow the input to fill the width of it's parent. Perhaps the govuk class being applied there would solve the issues you saw.

FarronF (approved):

AWildBugAppeared (approved):

PR #1556, opened 2021-09-10
VIH-8041 switch between original and filtered streams

shaed-parkar:

VIH-8041

### Change description ###

* switch between original and filtered streams

```
[ ] Yes
[x] No
```

### DIFF
--- a/VideoWeb/VideoWeb/ClientApp/src/app/services/media-stream.service.spec.ts
+++ b/VideoWeb/VideoWeb/ClientApp/src/app/services/media-stream.service.spec.ts
@@ -46,11 +46,10 @@ describe('MediaStreamService', () => {
         navigatorSpy = jasmine.createSpyObj<Navigator>([], ['mediaDevices']);
         getSpiedPropertyGetter(navigatorSpy, 'mediaDevices').and.returnValue(mediaDevicesSpy);
 
-        videoFilterStreamServiceSpy = jasmine.createSpyObj<VideoFilterService>([
-            'initFilterFromMediaStream',
-            'startFilteredStream',
-            'doesSupportVideoFiltering'
-        ]);
+        videoFilterStreamServiceSpy = jasmine.createSpyObj<VideoFilterService>(
+            ['initFilterFromMediaStream', 'startFilteredStream', 'doesSupportVideoFiltering'],
+            ['filterOn']
+        );
         const filterStreamTracks = [];
         filterStreamTracks.push(
             jasmine.createSpyObj<MediaStreamTrack>(['stop'])
@@ -148,6 +147,7 @@ describe('MediaStreamService', () => {
             const expectedStreamClone = spyOn(expectedStream, 'clone');
             expectedStreamClone.and.returnValue(expectedStream);
             mediaDevicesSpy.getUserMedia.and.resolveTo(expectedStream);
+            getSpiedPropertyGetter(videoFilterStreamServiceSpy, 'filterOn').and.returnValue(false);
 
             // Act
             let resultantStream = null;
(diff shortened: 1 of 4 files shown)


omio-ma (dismissed):

VideoWeb/VideoWeb/ClientApp/src/app/services/media-stream.service.ts
```diff
@@ -42,7 +42,7 @@ export class MediaStreamService {
         return from(this.navigator.mediaDevices.getUserMedia({ video: { deviceId: { exact: device.deviceId } } })).pipe(
             map(stream => {
                 const cloneStream = stream.clone();
-                if (this.videoFilterService.doesSupportVideoFiltering()) {
+                if (this.videoFilterService.doesSupportVideoFiltering() && this.videoFilterService.filterOn) {
```
FarronF, line 45: Could we get the unit tests for this service to cover the new cases?

VideoWeb/VideoWeb/ClientApp/src/app/services/media-stream.service.spec.ts
```diff
@@ -157,7 +157,7 @@ describe('MediaStreamService', () => {
             // Assert
             expect(mediaDevicesSpy.getUserMedia).toHaveBeenCalledWith(cameraConstraintBuilder(cameraDevice));
 
-            expect(resultantStream).toBe(filterStream);
+            expect(resultantStream).toBe(expectedStream);
```
FarronF, line 160: Has this changed because of the additional condition in the service?

VideoWeb/VideoWeb/ClientApp/src/app/services/media-stream.service.ts
```diff
@@ -42,7 +42,7 @@ export class MediaStreamService {
         return from(this.navigator.mediaDevices.getUserMedia({ video: { deviceId: { exact: device.deviceId } } })).pipe(
             map(stream => {
                 const cloneStream = stream.clone();
```
FarronF, line 44: I know this isn't part of this particular change but could we just return the stream rather than cloning it if we aren't going to filter it? Or is this needed either way?

VideoWeb/VideoWeb/ClientApp/src/app/shared/select-media-devices/select-media-devices.component.ts
```diff
@@ -62,6 +62,10 @@ export class SelectMediaDevicesComponent implements OnInit, OnDestroy {
         this.userMediaService.activeMicrophoneDevice$.pipe(takeUntil(this.destroyedSubject)).subscribe(microphoneDevice => {
             this.updateSelectedMicrophone(microphoneDevice);
         });
+
+        this.videoFilterService.onFilterChanged$.pipe(takeUntil(this.destroyedSubject)).subscribe(() => {
+            this.updateSelectedCamera(this.selectedCameraDevice);
```
FarronF, line 67: Add unit test for this.

VideoWeb/VideoWeb/ClientApp/src/app/services/media-stream.service.spec.ts
shaed-parkar: yes

VideoWeb/VideoWeb/ClientApp/src/app/services/media-stream.service.ts
shaed-parkar: done

VideoWeb/VideoWeb/ClientApp/src/app/services/media-stream.service.ts
shaed-parkar: I gave it a go without cloning and it seems to be okay. Updated the tests so it no longer expects clone to be called too

omio-ma (approved):

FarronF (approved):

PR #1558, opened 2021-09-13
VIH-8041 Self view to include selected background

shaed-parkar:

VIH-8041

### Change description ###

Camera stream to be cloned else the stream does not include the selected background outside of the select device component

```
[ ] Yes
[x] No
```


FarronF (approved):
Code looks fine but discuss with Rob if it is needed before merging.
https://hmcts-reform.slack.com/archives/C01GG92A39P/p1631529658194400

PR #1568, opened 2021-09-14
VIH-8207 vho cc copying wrong (hearing id) quicklink data

alberto-hortiguela:
https://tools.hmcts.net/jira/browse/VIH-8207

### Change description ###
copy-quick-link onchange binding Input on conference update

```
[ ] Yes
[x] No
```


VideoWeb/VideoWeb/ClientApp/src/app/vh-officer/copy-quick-link/copy-quick-link.component.ts
```diff
@@ -21,7 +21,12 @@ export class CopyQuickLinkComponent implements OnInit {
     async ngOnInit() {
         this.tooltip = this.translateService.instant('copy-quick-link.display-text');
         const response = await this.vhoQueryService.getConferenceByIdVHO(this.conferenceId);
-        this.hearingId = response && response.hearing_id;
+        this.hearingId = response.hearing_id;
+    }
+
+    async ngOnChanges(): Promise<void> {
+        const response = await this.vhoQueryService.getConferenceByIdVHO(this.conferenceId);
```
FarronF, line 28: Why do we need to call the API this in this component?

alberto-hortiguela: The HearingId is not present in the HearingSummary object we are passing from its parent component. We need to retrieve it.

FarronF: Ahhh. I see. I didn't notice that this was in the ngOnInit as well. Should this ever change though?

Also is there any chance we could add the hearing id to that model? It seems like it would simplify this a lot.

VideoWeb/VideoWeb/ClientApp/src/app/vh-officer/vho-hearing-list/vho-hearing-list.component.ts
```diff
@@ -49,7 +49,7 @@ export class VhoHearingListComponent {
         return this.mapToHearing(conference, participants);
     }
 
-    trackConference(conference: HearingSummary) {
+    trackConference(index: string, conference: HearingSummary) {
```
FarronF, line 52: index should be a number

VideoWeb/VideoWeb/ClientApp/src/app/vh-officer/copy-quick-link/copy-quick-link.component.ts
```diff
@@ -21,7 +21,7 @@ export class CopyQuickLinkComponent implements OnInit {
     async ngOnInit() {
         this.tooltip = this.translateService.instant('copy-quick-link.display-text');
         const response = await this.vhoQueryService.getConferenceByIdVHO(this.conferenceId);
-        this.hearingId = response && response.hearing_id;
+        this.hearingId = response.hearing_id;
```
FarronF, line 24: Maybe response?.hearing_id;

omio-ma, line 24: dont think this change is required?

VideoWeb/VideoWeb/ClientApp/src/app/vh-officer/copy-quick-link/copy-quick-link.component.ts
```diff
@@ -1,4 +1,4 @@
-import { Component, Input, OnInit } from '@angular/core';
+import { Component, Input, OnChanges, OnInit } from '@angular/core';
```
FarronF: Remove OnChanges

VideoWeb/VideoWeb/ClientApp/src/app/vh-officer/copy-quick-link/copy-quick-link.component.ts
FarronF: I'm actually not clear how the previous implementation worked. Should that not return a boolean?

alberto-hortiguela: added

alberto-hortiguela: I wondered how this could compile and return an Id...
to my understanding that should return a boolean too

FarronF (dismissed):

VideoWeb/VideoWeb/ClientApp/src/app/vh-officer/copy-quick-link/copy-quick-link.component.ts
omio-ma: no basically its the same as using an if block to check if response is null or not, if its not null then it will evaluate the next code basically which is similar to this:
`
if(response !== null){
this.hearingId = response.hearing_id
}
`

alberto-hortiguela: I have reverted to the original. Its the same behavior based in your investigation **response** && **response.hearing_id and response?.hearing_id**

omio-ma (approved):

VideoWeb/VideoWeb/ClientApp/src/app/vh-officer/copy-quick-link/copy-quick-link.component.ts
FarronF: This seems a very confusing way to write it though. I'll approve since it's not an actual but in the future I think we should use clearer patterns. Here is a stack overflow about it if interested.

https://stackoverflow.com/questions/12878612/assignment-with-double-ampersand/12878623

FarronF (approved):

PR #1580, opened 2021-09-20
VIH-7908 admit dismiss quick link users UI

FarronF:
VIH-7908
VIH-8235

### Change description ###
Updating judge UI for calling/admission/dismissing to use the hearing role rather than be hardcoded to witness.

```
[ ] Yes
[x] No
```


VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/participants-panel/participants-panel.component.html
```diff
@@ -119,7 +119,12 @@ <h2 class="govuk-heading-s white">{{ 'participants-panel.participants' | transla
               *ngIf="!participant.transferringIn"
               appTooltip
               [text]="
-                (participant.isAvailable() ? 'participants-panel.call-witness' : 'participants-panel.witness-unavailable') | translate
+                (participant.isAvailable() ? 
+                  (isWitness(participant) ? 
+                    ('participants-panel.call-witness' | translate) : 
+                    ('participants-panel.admit-participant' | translate : {role: 'hearing-role.' + (participant.hearingRole | hyphenate) | translate | lowercase})
+                  ) : 
+                  'participants-panel.participant-unavailable' | translate : {role: 'hearing-role.' + (participant.hearingRole | hyphenate) | translate})
```
FarronF, line 127: participant.isWitness would not work for me and always returned the admit-participant value. Not sure why. Any insight/advice/thoughts on this would be appreciated as I don't like having this additional method.

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/judge-context-menu/judge-context-menu.component.spec.ts
```diff
@@ -193,69 +250,85 @@ describe('JudgeContextMenuComponent', () => {
...
-        spyOn(elementRef.nativeElement, 'contains').and.returnValue(false);
-        const event = new MouseEvent('click');
-        spyOnProperty(event, 'target').and.returnValue(new EventTarget());
+    describe('isClickedOutsideOfOpenMenu', () => {
+        let event: MouseEvent;
+        let containsSpy: jasmine.Spy;
+        beforeEach(() => {
+            event = new MouseEvent('click');
+            spyOnProperty(event, 'target').and.returnValue(document.createElement('div'));
+            containsSpy = spyOn(component['elementRef'].nativeElement, 'contains');
```
FarronF, line 259: I really dislike spying on a private input but I couldn't get the elementRef spy that was injected to work. If anyone has any thoughts or ideas on another way to do this I'd be open to it.

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/participants-panel/participants-panel.component.html
FarronF: Although this has been done a similar way in other components.

VideoWeb/VideoWeb/ClientApp/src/assets/i18n/en.json
```diff
@@ -382,7 +382,8 @@
   "judge-context-menu": {
     "context-menu-icon": "Context menu icon",
     "call-witness": "Call witness",
-    "dismiss-witness": "Dismiss witness",
+    "admit-participant": "Admit {{ role }}",
```
FarronF, line 385: Ticket VIH-8253 pending for Welsh

VideoWeb/VideoWeb/ClientApp/src/app/shared/pipes/hyphenate.pipe.ts
```diff
@@ -0,0 +1,13 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
+    name: 'hyphenate',
+    pure: false
+})
+export class HyphenatePipe implements PipeTransform {
+    constructor() {}
```
omio-ma, line 8: remove empty constructor, don't think that will cause an issue?

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/judge-context-menu/judge-context-menu.component.html
```diff
@@ -1,13 +1,43 @@
...
-  <a *ngIf="!participant.isJudge && participant.hasHandRaised() && participant.isInHearing()" (click)="lowerParticipantHand()">{{'judge-context-menu.lower-hand' | translate}}</a>
-  <a *ngIf="participant.isInHearing() && !participant.isLocalCameraOff()" (click)="toggleSpotlightParticipant()">{{ getPinStatusText() }}</a>
+  <a *ngIf="participant.isInHearing()" (click)="toggleMuteParticipant()" [id]="getElementId('toggle-mute')">{{ getMutedStatusText() }}</a>
+  <a *ngIf="canCallParticipantIntoHearing()" (click)="callParticipantIntoHearing()" [id]="getElementId('call')">
+    <ng-container
+      *ngIf="isWitness; then isWitnessCallContent; else isNotWitnessCallContent">
+    </ng-container>
+  </a>
+  <a *ngIf="canDismissParticipantFromHearing()" (click)="dismissParticipantFromHearing()" [id]="getElementId('dismiss')">
+    {{'judge-context-menu.dismiss-participant' | translate: { role: 'hearing-role.' + (participant.hearingRole | hyphenate) | translate | lowercase} }}
```
FarronF, line 32: lowercase pipe here is not ideal, really we would like to have the roles in lowercase except for acronyms and the like. But it works for all the roles that will appear here.

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/participants-panel/participants-panel.component.html
omio-ma: I am not sure why that wouldn't work because if participant was null it would just show unavailable text instead. So does the unit test fail when you switch from using this function to particiapnt.isWitness? 

Also I was wondering if we could refactor this block completely into the typescript file, move all the logic inside of a function that returns the expected text instead as having if condition this way is a bit difficult to read. I am not sure if we get any benefits in having this all in the template file.

alberto-hortiguela:
8245 is on a separate PR https://github.com/hmcts/vh-video-web/pull/1583, is there something added in the last minute?

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/judge-context-menu/judge-context-menu.component.ts
```diff
@@ -85,27 +85,21 @@ export class JudgeContextMenuComponent {
         this.isDroppedDown = !this.isDroppedDown;
     }
 
-    getAdditionalText(): string {
-        return this.participant.hearingRole !== HearingRole.JUDGE ? this.getHearingRole() + this.getCaseRole() : '';
+    get isJudge(): boolean {
+        return this.participant.hearingRole === HearingRole.JUDGE;
```
alberto-hortiguela, line 89: DualHost role will have to be added same here mentioned by @krishnapottekkat-hmcts

VideoWeb/VideoWeb/ClientApp/src/assets/i18n/en.json
alberto-hortiguela: I am missing this part wow the role gets inserted here

VideoWeb/VideoWeb.UnitTests/Controllers/ConferenceManagement/DismissParticipantTests.cs
```diff
@@ -162,23 +190,25 @@ public async Task should_return_accepted_when_participant_is_a_room()
...
             var room = TestConference.CivilianRooms.First();
-            var witness = TestConference.Participants.First(x => x.IsWitness() && room.Participants.Contains(x.Id));
+            var participant = TestConference.Participants.First(x => x.IsWitness() && room.Participants.Contains(x.Id));
             var user = new ClaimsPrincipalBuilder()
                 .WithUsername(judge.Username)
                 .WithRole(AppRoles.JudgeRole).Build();
             Controller = SetupControllerWithClaims(user);
 
-            var result = await Controller.DismissParticipantAsync(TestConference.Id, witness.Id);
+            string expectedBody = $"{participant.HearingRole} dismissed by {judge.HearingRole}";
```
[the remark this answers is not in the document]
FarronF: Yes. Staff member could as well. Do you think that should be done in these tests?

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/judge-context-menu/judge-context-menu.component.ts
FarronF: I'm not sure how this will interact with staff member. As there was no case for this in the existing code I didn't add it here.

VideoWeb/VideoWeb/ClientApp/src/assets/i18n/en.json
FarronF: I can show you how this works. I think I have simpler examples somewhere, the one here ends up going though multiple pipes.

VideoWeb/VideoWeb/ClientApp/src/app/shared/pipes/hyphenate.pipe.ts
FarronF: Done

VideoWeb/VideoWeb.UnitTests/Controllers/ConferenceManagement/DismissParticipantTests.cs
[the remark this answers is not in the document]
FarronF: I looked at doing this. But the implementation for calling witnesses was not done so this would need to be done. I know the AC states dual host but is there other tickets for that functionality?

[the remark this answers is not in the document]
krishnapottekkat-hmcts: I believe the story mention that in the subject. I don't have an intention to block this ticket with Dual Host/Staff member is not released with 1.28 as well. So we can create another 1 point Story to cover Staff Member scenario so that it is tracked and we push forward with this story to QA. Hope that's fine @simonwkelso

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/judge-context-menu/judge-context-menu.component.ts
```diff
@@ -85,27 +85,21 @@ export class JudgeContextMenuComponent {
...
-
-    private showCaseRole() {
+    showCaseTypeGroup(): boolean {
         return !this.participant.caseTypeGroup ||
             this.participant.caseTypeGroup.toLowerCase() === CaseTypeGroup.NONE.toLowerCase() ||
             this.participant.caseTypeGroup.toLowerCase() === CaseTypeGroup.OBSERVER.toLowerCase() ||
             this.participant.caseTypeGroup.toLowerCase() === CaseTypeGroup.PANEL_MEMBER.toLowerCase() ||
             this.participant.caseTypeGroup.toLowerCase() === CaseTypeGroup.JUDGE.toLowerCase() ||
-            this.participant.caseTypeGroup.toLowerCase() === 'endpoint'
+            this.participant.caseTypeGroup.toLowerCase() === CaseTypeGroup.ENDPOINT.toLowerCase()
```
FarronF, line 102: This should also include staff member if we are handling that. @krishnapottekkat-hmcts @alberto-hortiguela @simonwkelso

VideoWeb/VideoWeb.UnitTests/Controllers/ConferenceManagement/DismissParticipantTests.cs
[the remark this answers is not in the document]
FarronF: As discussed this will be covered by VIH-8131

VideoWeb/VideoWeb.UnitTests/Controllers/ConferenceManagement/DismissParticipantTests.cs
```diff
@@ -137,23 +137,87 @@ public async Task should_return_accepted_when_participant_is_witness_and_judge_i
...
+                case Role.Judge:
+                    appRole = AppRoles.JudgeRole.ToString();
+                    break;
+                case Role.StaffMember:
+                    appRole = AppRoles.StaffMember.ToString();
+                    break;
+                default:
+                    appRole = "";
+                    break;
+            }
```
FarronF, line 204: Wanted to have this as part of the test case but couldn't as they are static. I left StaffMember even though the test cases doesn't cover them just for ease when they are added back. 
Does this make sense?

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/judge-context-menu/judge-context-menu.component.ts
FarronF: As discussed this will be covered by VIH-8131

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/participants-panel/participants-panel.component.html
FarronF: It's preferred to use pipes for a few reasons. Performance is one. If the pipes are pure they will only be called when the input changes as opposed to every change detection cycle. I don't know for sure if TranslatePipe is pure but the other 2 are.
 
The other is if we have it in the typescript file we need to worry about subscribing and unsubscribing from different observables. By using pipes that is all taken care of for us.

FarronF: I will say the layout here isn't great. Since it's a directive input I'm not sure if there's a nicer way to do it. I don't think a function will work as the pipes vary for each case, but I'll have a little think.

krishnapottekkat-hmcts (approved):
LGTM

VideoWeb/VideoWeb.UnitTests/Controllers/ConferenceManagement/DismissParticipantTests.cs
omio-ma: instead of using the switch cases can we not just include a second argument in the test case and pass app role along with dismisserRole, ie, like this:
      [Test]
        [TestCase(Role.Judge, "JudgeRole")]
        public async Task should_create_an_alert_with_the_correct_dismisser_role_when_the_witness_is_dismissed(Role dismisserRole, string appRole)

FarronF: It becomes a magic string issue then though. Although this switch statement feels a bit arcane itself.

FarronF: Agh. Actually I think earlier I was trying to use it like an enum instead of a string. Which makes sense. I'll have a nicer solution next commit.

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/participants-panel/participants-panel.component.html
FarronF: I'm afraid I can't come up with a nicer way to do this as the only way I can think would be to use *ngIf with ng-templates but that would involve duplicating the other attributes which I think would not be worth it.

omio-ma (approved):

PR #1582, opened 2021-09-21
VIH-7871 enable mobile support

AWildBugAppeared:
https://tools.hmcts.net/jira/browse/VIH-7871


VideoWeb/VideoWeb/ClientApp/src/app/services/device-type.service.ts
```diff
@@ -28,6 +28,14 @@ export class DeviceTypeService {
         );
     }
 
+    isIOS(): boolean {
+        return this.deviceDetectorService.os.toLowerCase() === 'ios';
```
robertbennett1998, line 32: Could use the OS const from node_modules\ngx-device-detector\lib\device-detector.constants.d.ts

robertbennett1998 (dismissed):
LGTM

robertbennett1998 (approved):
LGTM

azure-pipelines.beta.yml
```diff
@@ -74,19 +74,19 @@ variables:
...
 trigger: none
 pr: none
 schedules:
-- cron: "0 19 * * 0"
-  displayName: Weekly Beta build
-  always: true
-  branches:
-    include:
-    - master
+  - cron: '0 19 * * 0'
```
FarronF, line 84: A little concerned about this and the spacing changes in the yml files. Aren't they very spacing sensitive or am I mistaken? I will approve and let the pipelines determine that though.

VideoWeb/VideoWeb/ClientApp/src/app/services/device-type.service.spec.ts
```diff
@@ -97,6 +97,22 @@ describe('DeviceType', () => {
         expect(service.getOSVersion()).toBe(testOsVersion);
     });
 
+    it('should return whether hte OS is iOS', () => {
```
FarronF, line 100: Minor. Misspelling of "the" don't worry about it if there's no other changes to make

VideoWeb/VideoWeb/ClientApp/src/app/home/navigator/navigator.component.ts
```diff
@@ -5,28 +5,42 @@ import { DeviceTypeService } from '../../services/device-type.service';
...
     ngOnInit() {
-        if (this.deviceTypeService.isDesktop() || this.deviceTypeService.isIpad()) {
-            this.profileService
-                .getUserProfile()
-                .then(profile => this.navigateToHearingList(profile))
-                .catch(error => this.errorService.handleApiError(error));
-        } else {
-            this.router.navigate([pageUrls.UnsupportedDevice]);
-        }
+        this.configService.getClientSettings().subscribe(settings => {
```
FarronF, line 27: Shouldn't we unsubscribe from this on destroy?

VideoWeb/VideoWeb/ClientApp/src/app/home/navigator/navigator.component.ts
```diff
@@ -5,28 +5,42 @@ import { DeviceTypeService } from '../../services/device-type.service';
...
+            this.enableAndroidSupport = settings.enable_android_support;
```
omio-ma, line 28: we should just use the value from the settings as its not used anywhere else

VideoWeb/VideoWeb/ClientApp/src/app/home/navigator/navigator.component.ts
FarronF: Oh is this an API call that will resolve itself? I would prefer to be more specific here but I'll approve if it is.

FarronF (approved):

azure-pipelines.yml
```diff
@@ -323,6 +356,35 @@ extends:
       StrykerBreak: 75
     ACTest: ${{ parameters.RunACTests }}
     releaseParameters:
+      variables:
+        - name: EnableAndroidSupport
+          value: false
```
alberto-hortiguela, line 361: Why are we setting the values in the yml files instead of using the variables from the group?

alberto-hortiguela (approved):
LGTM

PR #1583, opened 2021-09-21
VIH-8245 witnesses displayed as Available when in the waiting room

alberto-hortiguela:
https://tools.hmcts.net/jira/browse/VIH-8245

### Change description ###
To display "Available" tooltip message when a witness on the WR is callable

```
[ ] Yes
[x] No
```
image


MichaelKeaneathmcts:
This might need an automation test at some point.  Is there a unique reference to the web element?  I'm looking for code such as <... id={participantId}-element-name ...>

Any automation tests can be brittle unless there is a very clear reference to an element, searching the DOM can mean that the next web change breaks the tests.

FarronF:
> 
> 
> This might need an automation test at some point. Is there a unique reference to the web element? I'm looking for code such as <... id={participantId}-element-name ...>
> 
> Any automation tests can be brittle unless there is a very clear reference to an element, searching the DOM can mean that the next web change breaks the tests.

@MichaelKeaneathmcts Maybe we should discuss a way we could approach this in the future. I'm ensuring a component I'm working on now has ids as such `judge-context-menu-participant-${participantId}-${elementId}` but all of this is pretty adhoc and we don't have an easy way to generate these ids. And I don't think we have a clear style for what these ids should look like.

FarronF (changes requested):
Looking at Lorenzo's comment on the ticket it states the following cases

    Joining
    Available
    In consultation
    Disconnected
    Not signed in

I don't see "In consultation" covered here. We should implement that as part of this.

alberto-hortiguela:
> Looking at Lorenzo's comment on the ticket it states the following cases
> 
> ```
> Joining
> Available
> In consultation
> Disconnected
> Not signed in
> ```
> 
> I don't see "In consultation" covered here. We should implement that as part of this.

You are right. However should not be then a ticket ( with points ) instead of a bug fix? This looks like new functionality

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/participants-panel/participants-panel.component.ts
```diff
@@ -506,16 +506,27 @@ export class ParticipantsPanelComponent implements OnInit, OnDestroy {
...
+                participant.displayName + ': ' + this.getTranslatedText('participant-available') + this.getAdditionalText(participant);
         }
         if (participant.isDisconnected()) {
-            return participant.displayName + ': ' + this.getTranslatedText('disconnected') + this.getAdditionalText(participant);
+            toolTipText = participant.displayName + ': ' + this.getTranslatedText('disconnected') + this.getAdditionalText(participant);
         }
-        return participant.displayName + this.getAdditionalText(participant);
+
+        if (!toolTipText) {
+            toolTipText = participant.displayName + this.getAdditionalText(participant);
```
omio-ma, line 526: we can use this to initialize the toolTipText property, removes the need for this if block

FarronF:
> 
> 
> > Looking at Lorenzo's comment on the ticket it states the following cases
> > ```
> > Joining
> > Available
> > In consultation
> > Disconnected
> > Not signed in
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I don't see "In consultation" covered here. We should implement that as part of this.
> 
> Is there a way to have a hearing in session and participants having a private consultation?
> If we are bringing new functionality should we not create a ticket ( with points ) instead of a bug fix one?

Yeah I think you're right actually. This case either was missed as an AC or was not considered. Lets discuss this with the team tomorrow.

omio-ma:
> This might need an automation test at some point. Is there a unique reference to the web element? I'm looking for code such as <... id={participantId}-element-name ...>
> 
> Any automation tests can be brittle unless there is a very clear reference to an element, searching the DOM can mean that the next web change breaks the tests.

I had to check the front end code for this, currently the tooltip is an angular directive, and it doesn't accept a property that can be used to dynamically create a unique id. It is do-able but I believe that should be part of another ticket as it would impact all the other tooltips we have in our application. Also I don't know how easily we can do an automation test for this, because this tooltip is actually not being hidden instead being pushed out off screen (ie, its using top and left css properties to take it off display) which would mean we cant just rely on asserting that the tooltip element is visible when hovered over the judge's name

MichaelKeaneathmcts:
Thanks for the feedback, in this case then an ID is probably not required.

I hope that everyone will think about what elements can be referenced by automation tests going forward.

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/participants-panel/participants-panel.component.ts
alberto-hortiguela: that's done now

omio-ma (approved):

krishnapottekkat-hmcts (approved):

FarronF (approved):

PR #1593, opened 2021-09-23
VIH-8250 add unique id's to quicklink web elements for automation

alberto-hortiguela:
https://tools.hmcts.net/jira/browse/VIH-8250

### Change description ###
Components updated to include non-static property IDs on quick link web elements

```
[ ] Yes
[x ] No
```


VideoWeb/VideoWeb/ClientApp/src/app/vh-officer/copy-id/copy-id.component.ts
```diff
@@ -9,11 +9,13 @@ import { ClipboardService } from 'ngx-clipboard';
 export class CopyIdComponent implements OnInit {
     @Input() conference: HearingSummary;
     tooltip: string;
+    propertyIdName = 'copy-conference-id';
     @ViewChild('copyID', { static: false }) copyID: ElementRef;
 
     constructor(private clipboardService: ClipboardService) {}
 
     ngOnInit(): void {
+        this.propertyIdName = this.propertyIdName + '-' + this.conference.id;
```
FarronF, line 18: There is a slight risk here that if the inputs changed but the component stayed that the id would be wrong. This could happen in the case where we saw the wrong link being copied when one was removed.

I was doing the same and was considering using ngOnChanges instead. I don't think we need to be too concerned with this risk here and if that happened it would signal that there is something more fundamentally wrong with our code. So I think we are fine to leave it as it is.

Just be careful in tests if you call ngOnInit this could append the id multiple times.

VideoWeb/VideoWeb/ClientApp/src/app/vh-officer/copy-quick-link/copy-quick-link.component.spec.ts
```diff
@@ -47,13 +47,15 @@ describe('CopyQuickLinkComponent', () => {
     });
 
     it('renders icon to copy to clipboard', () => {
+        component.ngOnInit();
```
FarronF, line 50: Could you check if this causes the Id to have 2 ids?

VideoWeb/VideoWeb/ClientApp/src/app/vh-officer/copy-quick-link/copy-quick-link.component.spec.ts
```diff
@@ -47,13 +47,15 @@ describe('CopyQuickLinkComponent', () => {
...
+        fixture.detectChanges();
         expect(fixture.debugElement.query(By.css('#copy-quick-link'))).toBeTruthy();
     });
 
     it('copies content into the clipboard using the clipboard service when element is clicked', fakeAsync(() => {
         const baseUrl = 'https://wow';
         spyOn(component, 'getbaseUrl').and.returnValue(baseUrl);
-        const copyLinkElement = fixture.debugElement.query(By.css('#copy-quick-link'));
+        const copyLinkElement = fixture.debugElement.query(By.css('span'));
```
FarronF, line 58: Perhaps we could search for the expected id to ensure it is being set properly. Or maybe even have a separate test.

VideoWeb/VideoWeb/ClientApp/src/app/vh-officer/copy-quick-link/copy-quick-link.component.spec.ts
alberto-hortiguela: Would this be valid?
image

VideoWeb/VideoWeb/ClientApp/src/app/vh-officer/copy-id/copy-id.component.ts
alberto-hortiguela: Would that be related to the same scenario we got where the trackedBy on that for loop was creating issues?

VideoWeb/VideoWeb/ClientApp/src/app/vh-officer/copy-quick-link/copy-quick-link.component.spec.ts
```diff
@@ -47,13 +47,15 @@ describe('CopyQuickLinkComponent', () => {
     });
 
     it('renders icon to copy to clipboard', () => {
+        component.ngOnInit();
+        fixture.detectChanges();
```
FarronF, line 51: I don't think these lines are needed since it's done in the before each. Was there a reason to add them?

omio-ma (approved):

MichaelKeaneathmcts (approved):

PR #1595, opened 2021-09-24
VIH-8235 case type group translation

FarronF:
VIH-8235

### Change description ###
Adding translation to case type group/party for judge context menu. Also changing the way ids for elements are set.

```
[ ] Yes
[x] No
```


VideoWeb/VideoWeb.UnitTests/Controllers/ConferenceManagement/DismissParticipantTests.cs
```diff
@@ -137,23 +137,74 @@ public async Task should_return_accepted_when_participant_is_witness_and_judge_i
...
+            typedResult.Should().NotBeNull();
+
+
+            VideoApiClientMock.Verify(x => x.AddTaskAsync(TestConference.Id,
+                It.Is<AddTaskRequest>(r => r.ParticipantId == participant.Id && r.Body == expectedBody && r.TaskType == TaskType.Participant)),
+                Times.Once);
+        }
+
         [Test]
         public async Task should_create_an_alert_when_the_witness_is_dismissed()
```
alberto-hortiguela, line 166: As we are dismissing participants ( including QL paraticipants ) should we reflect this in the unit test name?

PR #1607, opened 2021-09-29
VIH-8279 Fixed microphone width and self test misalignment

AWildBugAppeared:
https://tools.hmcts.net/jira/browse/VIH-8279

### DIFF
--- a/VideoWeb/VideoWeb/ClientApp/src/app/shared/self-test/self-test.component.scss
+++ b/VideoWeb/VideoWeb/ClientApp/src/app/shared/self-test/self-test.component.scss
@@ -7,3 +7,8 @@
   color: govuk-colour('light-blue');
   text-decoration-line: underline;
 }
+
+.self-view-container {
+  width: 270px;
+  padding: 0 15px;
+}
--- a/VideoWeb/VideoWeb/ClientApp/src/app/shared/self-test/self-test.component.html
+++ b/VideoWeb/VideoWeb/ClientApp/src/app/shared/self-test/self-test.component.html
@@ -15,7 +15,7 @@
 </div>
 
 <div class="govuk-grid-row">
-  <div>
+  <div class="self-view-container">
     <div>
       <video
         *ngIf="streamsActive"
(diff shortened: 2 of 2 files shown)


alberto-hortiguela (dismissed):
LGTM

ZJianWang (commented):
looks good to me but microphone bar still longer than video, did you talk to Neil, is Neil happy with change?

AWildBugAppeared:
> looks good to me but microphone bar still longer than video, did you talk to Neil, is Neil happy with change?

AWildBugAppeared:
The AC was quite clear in that it should be half the current length so I've halved it. I would only ask Neil if the actual behaviour was unclear on the ticket itself. If the reporter however hasn't checked with Neil and this turns out to be incorrect, we can flag this in the next retro to make sure UI changes are correct when making the ticket.

premk-hmcts (commented):
self-view_preprod

abiodunoke:
> https://tools.hmcts.net/jira/browse/VIH-8279
> 
> image

FarronF (changes requested):
As the video above is not responsive and is just a set 270px I think this should just match the width for that. I don't think using the half width column would behave the way it does on preprod.

VideoWeb/VideoWeb/ClientApp/src/app/shared/self-test/self-test.component.scss
```diff
@@ -7,3 +7,8 @@
   color: govuk-colour('light-blue');
   text-decoration-line: underline;
 }
+
+.self-view-container {
+  width: 270px;
```
FarronF, line 12: We now have this width defined in 2 places if it ever needs to be changed. Maybe setting the element width in the html to 100% would help?

FarronF (approved):
I left a comment regarding the video width. However this page is not built in a responsive way in general and that should probably be done as part of a ticket to make it responsive.

VideoWeb/VideoWeb/ClientApp/src/app/shared/self-test/self-test.component.scss
AWildBugAppeared: I have fixed this to be 100%. Didn't actually think it would work before which is why I left it. Good shout

FarronF:
Images from running locally
image
image

FarronF (approved):

ZJianWang (approved):
LGTM

PR #1611, opened 2021-09-29
VIH-8128 updated toaster notifications for mobile tablet responsiveness

alberto-hortiguela:
https://tools.hmcts.net/jira/browse/VIH-8128

### Change description ###
Updated: Responsive Toastr notifications
image

image

```
[ ] Yes
[x] No
```


VideoWeb/VideoWeb/ClientApp/src/sass/vendor-modifications/_ngx-toastr-modifications.scss
```diff
@@ -0,0 +1,13 @@
+.toast-top-right {
+    top: 70px;
+    right: 20px;
+}
+
+.toast-top-center {
+    top: 70px !important;
+    @media only screen and (min-width: $breakpoint-medium) {
+        top: 70px !important;
```
omio-ma, line 9: just a query, did we get this style from Neil?

FarronF:
The effect around the toastrs looks a bit odd to me. That is the style that we had before but in the mockup it looks like it has a dark shadow rather that a light one. I can't really make a style call so I think you should arrange a call with Neil to discuss this.

FarronF (changes requested):
I'm afraid this doesn't match the mock up. The button should be full width on a smaller screen. But also the AC and the mockup seems to conflict when it comes to centering. Discuss with Neil and see if we can get some more defined numbers for widths and the like.

Let me know if you want me to join any discussion as well and I can make time.

VideoWeb/VideoWeb/ClientApp/src/sass/vendor-modifications/_ngx-toastr-modifications.scss
alberto-hortiguela: We imported the properties from the already toastr global styles we have in the code base

alberto-hortiguela:
This tikcet is moved to next sprint. 
This remote branch will be available to be merged once https://tools.hmcts.net/jira/browse/VIH-8143 is in Master

PR #1636, opened 2021-10-08
VIH-8328 VMR participants could be seen in hearing and consultations when camera is switched off from WR

robennett2:
[VIH-8328](https://tools.hmcts.net/jira/browse/VIH-8328)

### Change description ###
- Replace active camera track with an audio only image track when audio only is turned on rather than removing the track

```
[ ] Yes
[x] No
```

### DIFF
--- a/VideoWeb/VideoWeb/ClientApp/src/app/services/device-type.service.spec.ts
+++ b/VideoWeb/VideoWeb/ClientApp/src/app/services/device-type.service.spec.ts
@@ -130,7 +130,17 @@ describe('DeviceType', () => {
 
     isSupportedBrowserTestCases.forEach(test => {
         it(`should return ${test.expected} when browser is ${test.browser}`, () => {
+            deviceDetectorService.isDesktop.and.returnValue(true);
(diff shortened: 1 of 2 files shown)


VideoWeb/VideoWeb/ClientApp/angular.json
```diff
@@ -1,155 +1,176 @@
...
-            ],
-            "aot": true
-          },
-          "configurations": {
-            "production": {
-              "fileReplacements": [
-                {
-                  "replace": "src/environments/environment.ts",
-                  "with": "src/environments/environment.prod.ts"
+    "$schema": "./node_modules/@angular/cli/lib/config/schema.json",
```
omio-ma, line 2: are there any changes made in this json file or just format change ?

PrasannaDommalapati (approved):
LGTM

VideoWeb/VideoWeb/ClientApp/angular.json
robertbennett1998: There is a temporary unit test exclusion but the rest is formatting.

PR #1646, opened 2021-10-11
VIH-8340 Wrong camera is selected on refresh if user had selected none default camera

robennett2:
[VIH-8340](https://tools.hmcts.net/jira/browse/VIH-8340)

### Change description ###
- Fixed an issue where the wrong camera was loading as user_media_stream wasn't been set before pexip was ready which let it select default devices.

```
[ ] Yes
[x] No
```


alberto-hortiguela (approved):
LGTM

PR #1695, opened 2021-11-04
VIH-7768 display crest based on hearing venue

omio-ma:
https://tools.hmcts.net/jira/browse/VIH-7768

### Change description ###
- updated video api nuget package reference to 1.28.1
- scottish crest logo to be displayed for scottish hearing venue 

```
[ ] Yes
[x] No
```
consultation room participant panel:
consultation room participant panel

footer logo
footer logo

footer mobile
footer mobile

header logo mobile
header logo mobile

header-footer-participant-panel-tablet
header-footer-participant-panel-tablet

hearing room participant panel
hearing room participant panel

hearing rules mobile
hearing rules mobile

hearing rules tablet
hearing rules tablet

introduction mobile
introduction mobile

introduction tablet
introduction tablet

mobile participant panel
mobile participant panel

participant declaration logo
participant declaration logo

participant introduction logo
participant introduction logo

participant panel in hearing mobile
participant panel in hearing mobile

participant panel in hearing tablet
participant panel in hearing tablet

waiting room logo
waiting room logo

panel member in participants panel
panel member in participants panel

panel member in participants panel -mobile
panel member in participants panel -mobile

panel member in participants panel - tablet
panel member in participants panel - tablet


shaed-parkar (commented):
I can't help but think a directive would have been perfect here. It could have swapped between the different crests and an attribute for the the size would have been very clean

omio-ma:
@shaed-parkar  I have raised that with few other devs regarding if I could implement and use directive here, but the way some of the components are implemented (like images are sometimes background images) and the random nature of which image needs to be used would end up having a really complicated directive and wont be something reusable.

shaed-parkar:
> @shaed-parkar I have raised that with few other devs regarding if I could implement and use directive here, but the way some of the components are implemented (like images are sometimes background images) and the random nature of which image needs to be used would end up having a really complicated directive and wont be something reusable.

No harm in having multiple directives in my opinion. The components are doing more than they need to and personally should not have to track if IsScottish across each view. The first `S` in `SOLID` would totally allow multiple directives since they have a slight difference. Alternatively you have a single directive that supports attributes so that all the logic to decide which image to return is held in one place

omio-ma:
> > @shaed-parkar I have raised that with few other devs regarding if I could implement and use directive here, but the way some of the components are implemented (like images are sometimes background images) and the random nature of which image needs to be used would end up having a really complicated directive and wont be something reusable.
> 
> No harm in having multiple directives in my opinion. The components are doing more than they need to and personally should not have to track if IsScottish across each view. The first `S` in `SOLID` would totally allow multiple directives since they have a slight difference. Alternatively you have a single directive that supports attributes so that all the logic to decide which image to return is held in one place

so if I go with multiple directive, I would have a directive for each of those changes, the only place the image was similar is when I used the same image for the participant panel and the footer, however the footer is using a background image. So that would mean I will have a directive per component really as they are using different images. 

With a single directive however, the only way I can see something implemented would be where I can flag out which image to use which I couldnt envision being something nice, thats when I thought using directive wont work out here.

VideoWeb/VideoWeb/ClientApp/src/app/on-the-day/hearing-rules/hearing-rules.component.ts
```diff
@@ -1,29 +1,42 @@
...
+        private hearingVenueFlagsService: HearingVenueFlagsService
     ) {
         super(participantStatusUpdateService, logger);
     }
 
     ngOnInit() {
         this.conferenceId = this.route.snapshot.paramMap.get('conferenceId');
+
+        this.hearingVenueFlagsServiceSubscription$ = this.hearingVenueFlagsService.hearingVenueIsScottish$.subscribe(
+            value => (this.hearingVenueIsInScotland = value)
```
FarronF, line 34: This would actually be a good place to use the async pipe which should be preferred for these cases then we don't need to handle the subscription and disposing it.

https://angular.io/api/common/AsyncPipe

omio-ma: how would I use the async pipe in the template, I am using the flag hearingVenueIsInScotland in the template and not the hearingVenueFlagsServiceSubscription$

PrasannaDommalapati:
@omio-ma could you add a panel member/winger to the Scottish venue hearing and see it displays the correct crest next to that participant and add a screenshot also to this PR, please.

omio-ma:
> @omio-ma could you add a panel member/winger to the Scottish venue hearing and see it displays the correct crest next to that participant and add a screenshot also to this PR, please.

@PrasannaDommalapati  added the images titled: "panel member in participants panel"

VideoWeb/VideoWeb/ClientApp/src/app/on-the-day/judge-hearing-list/judge-hearing-table.component.ts
```diff
@@ -20,12 +21,12 @@ export class JudgeHearingTableComponent implements OnInit {
...
     @Output() selectedConference = new EventEmitter<ConferenceForHostResponse>();
 
-    constructor(private logger: Logger) {}
+    constructor(private logger: Logger, private hearingVenueFlagsService: HearingVenueFlagsService) {}
 
     ngOnInit() {
         this.hearings = this.conferenceForHostResponse.map(c => new JudgeHearingSummary(c));
         const last = this.hearings.pop();
-
+        this.hearingVenueFlagsService.setHearingVenueIsScottish(false);
```
FarronF, line 29: Any reason this can't go in the judge-hearing-list.component?

VideoWeb/VideoWeb/ClientApp/src/app/services/hearing-venue-flags.service.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+            done();
+        });
+    });
+
+    it('setHearingVenueIsScottish updates current value', done => {
+        service.setHearingVenueIsScottish(true);
+        service.hearingVenueIsScottish$.subscribe(response => {
+            expect(response).toBe(true);
+            done();
+        });
```
FarronF, line 25: You could put this above line 21 just to confirm that a subscription will update after a change.

VideoWeb/VideoWeb/ClientApp/src/app/shared/footer/footer.component.scss
```diff
@@ -3,24 +3,37 @@
...
   white-space: nowrap;
 }
+
+.govuk-footer__copyright-logo {
+  background-image: url('/assets/images/govuk-crest.png');
+}
+
+.govuk-footer__copyright-logo-scottish {
+  background-image: url('/assets/images/scottish-crest-footer.png');
+}
```
FarronF, line 25: I think we should keep .govuk-footer__copyright-logo and always set the class to that. But have two new classes one of which being .govuk-footer__copyright-logo-scottish and something else for the other. And have that conditionally set with with the ngClass similar to what you've done

VideoWeb/VideoWeb/ClientApp/src/app/shared/footer/footer.component.scss
```diff
@@ -3,24 +3,37 @@
...
+
 @media only screen and (-webkit-min-device-pixel-ratio: 2),
   only screen and (min-resolution: 192dpi),
   only screen and (min-resolution: 2dppx) {
   .govuk-footer__copyright-logo {
     background-image: url('/assets/images/govuk-crest-2x.png');
   }
+
+  .govuk-footer__copyright-logo-scottish {
+    background-image: url('/assets/images/scottish-crest-footer.png');
```
FarronF, line 35: Did you confirm with Neil that this image is ok? If you make the change I suggested above you could also take this out.

VideoWeb/VideoWeb/ClientApp/src/app/on-the-day/judge-hearing-list/judge-hearing-table.component.ts
omio-ma: no, i cant remember why I put it in the hearing table component, for the partcipants it appears to be in the table component, so I can try to do that and see if it works as expected

VideoWeb/VideoWeb/ClientApp/src/app/services/hearing-venue-flags.service.spec.ts
omio-ma: yea cool I will add that assertion in all the tests

VideoWeb/VideoWeb/ClientApp/src/app/shared/footer/footer.component.scss
omio-ma: yea sure, so I will add a class called  govuk-footer-background-uk and govuk-footer-background-scottish

VideoWeb/VideoWeb/ClientApp/src/app/shared/footer/footer.component.scss
omio-ma: yea I confirmed with Neil, at the moment we dont have a high res image so asked me to use the same image. I would still need this change right because otherwise in high res it would then show the uk crest?

VideoWeb/VideoWeb/ClientApp/src/app/shared/header-logo-svg/header-logo-svg.component.html
```diff
@@ -1,4 +1,6 @@
-<svg role="presentation" focusable="false" class="moj-header__logotype-crest" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 40 40" height="40" width="40">
+<img role="presentation" id="header-logo-scot" class="moj-header__logotype-crest" *ngIf="hearingVenueIsInScotland" src="/assets/images/scot-crest-white.png" height="40" [attr.alt]="'participants-panel.scottish-crest' | translate"/>
+
+<svg id="header-logo-uk" *ngIf="!hearingVenueIsInScotland" role="presentation" focusable="false" class="moj-header__logotype-crest" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 40 40" height="40" width="40">
```
FarronF, line 3: I don't think we should put the img with png inside svg component. Maybe do this check in the header? Also if you have *ngIf for a true and false I would recommend using ng-template with *ngIf then else
https://ultimatecourses.com/blog/angular-ngif-else-then

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/judge-participant-status-list/judge-participant-status-list.component.html
```diff
@@ -3,11 +3,11 @@
     <div class="govuk-!-margin-top-4">
       <div class="govuk-grid-column-one-third">
         <img
+          [src]="(hearingVenueIsInScotland$ | async) ? '/assets/images/scottish-crest-footer.png': '/assets/images/govuk-crest.png'"
```
FarronF, line 6: Should it use the image for the footer here? We should rename that if it's to be used other places.

VideoWeb/VideoWeb/ClientApp/src/app/shared/header-logo-svg/header-logo-svg.component.html
omio-ma: yea I will move it out of the svg component

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/judge-participant-status-list/judge-participant-status-list.component.html
omio-ma: yea I will rename that to better reflect that image being used

VideoWeb/VideoWeb/ClientApp/src/app/shared/header-logo-svg/header-logo-svg.component.html
omio-ma: argh, also appears I have missed converting it into using an async pipe anyways

VideoWeb/VideoWeb/ClientApp/src/app/services/hearing-venue-flags.service.spec.ts
omio-ma: I just noticed this is in the actual service, if you see the previous test actually tests it being set to false by default, so not adding this change

FarronF:
I think it would be nice to have a component or directive to handle a lot of this. Especially now seeing how many places it's changed. But I'm not clear on how many different versions of the logo there are. I know the header and footer are outliers in that they use svg and background image respectively, they could be left out.

A directive should allow keeping the current classes as is it could just handle setting the src and alt.
https://stackoverflow.com/questions/53636835/how-can-i-get-an-image-src-value-using-a-directive-in-angular-2

But this might be considered too much refactoring of the current implementation, so I'm ok with either way.

VideoWeb/VideoWeb/ClientApp/src/app/shared/footer/footer.component.scss
FarronF: Would hmcts make more sense than uk? Or is it also hmcts in scotland?

VideoWeb/VideoWeb/ClientApp/src/app/shared/footer/footer.component.scss
FarronF: If you make the previous change I suggested so the background image is set by govuk-footer-background-uk or whatever you call it then it wouldn't for scottish ones.

VideoWeb/VideoWeb/ClientApp/src/app/on-the-day/hearing-rules/hearing-rules.component.ts
shaed-parkar: that's why I suggested creating a directive that supports inputs. I'm not going to stop you on this, just something to think about

VideoWeb/VideoWeb/ClientApp/src/app/shared/footer/footer.component.scss
omio-ma: oh that makes sense, no for scottish its SCTS, I will rename it

VideoWeb/VideoWeb/ClientApp/src/app/shared/footer/footer.component.html
```diff
@@ -36,7 +36,7 @@
             <div class="govuk-footer__meta-item">
                 <div>
                     <a class="link-container" aria-label="'footer.copyright-info-link-label' | translate" href="https://www.nationalarchives.gov.uk/information-management/re-using-public-sector-information/uk-government-licensing-framework/crown-copyright/">
-                        <span [ngClass]="(hearingVenueIsInScotland$ | async) ? 'govuk-footer__copyright-logo-scottish' : 'govuk-footer__copyright-logo'" role="presentation"></span>
+                        <span [ngClass]="(hearingVenueIsInScotland$ | async) ? 'govuk-footer__copyright-logo govuk-footer-background-scottish' : 'govuk-footer__copyright-logo govuk-footer-background-uk'" role="presentation"></span>
```
FarronF:
```suggestion
                        <span class="govuk-footer__copyright-logo" [ngClass]="(hearingVenueIsInScotland$ | async) ? 'govuk-footer-background-scottish' : 'govuk-footer-background-uk'" role="presentation"></span>
```

VideoWeb/VideoWeb/ClientApp/src/app/shared/footer/footer.component.scss
```diff
@@ -16,23 +16,23 @@
...
+  background-image: url('/assets/images/scottish-crest-grey.png');
 }
 
 @media only screen and (-webkit-min-device-pixel-ratio: 2),
   only screen and (min-resolution: 192dpi),
   only screen and (min-resolution: 2dppx) {
-  .govuk-footer__copyright-logo {
+  .govuk-footer-background-uk {
     background-image: url('/assets/images/govuk-crest-2x.png');
   }
```
FarronF: You should be able to nest this media query above line 21 just to structure it better. And remove the class here.

VideoWeb/VideoWeb/ClientApp/src/app/shared/footer/footer.component.scss
```diff
@@ -16,23 +16,23 @@
...
 
-  .govuk-footer__copyright-logo-scottish {
-    background-image: url('/assets/images/scottish-crest-footer.png');
+  .govuk-footer-background-scottish {
+    background-image: url('/assets/images/scottish-crest-grey.png');
   }
```
FarronF: This can be removed

VideoWeb/VideoWeb/ClientApp/src/app/shared/header/header.component.html
```diff
@@ -4,7 +4,11 @@
 
     <div class="moj-header__logo">
 
-    <app-header-logo-svg></app-header-logo-svg>
+    <img role="presentation" id="header-logo-scot" class="moj-header__logotype-crest" *ngIf="hearingVenueIsScottish$ | async;else ukCrestSvg" src="/assets/images/scot-crest-white.png" height="40" [attr.alt]="'participants-panel.scottish-crest' | translate"/>
```
FarronF, line 7: Could you move the img inside another template and put the ngIf inside a ng-container like this
```suggestion
    <ng-container
  *ngIf="hearingVenueIsScottish$ | async; then scottishCrest; else ukCrest">
</ng-container>
```

I feel otherwise the logic gets a bit lost amongst the other attributes.

VideoWeb/VideoWeb/ClientApp/src/app/shared/header/header.component.html
```diff
@@ -4,7 +4,11 @@
...
+
+    <ng-template #ukCrestSvg>
```
FarronF, line 9:
```suggestion
    <ng-template #ukCrest>
```
We don't want to give the impression that this has to be an svg

VideoWeb/VideoWeb/ClientApp/src/app/on-the-day/participant-hearings/hearing-list-table.component.spec.ts
```diff
@@ -7,13 +10,24 @@ describe('HearingListTableComponent', () => {
     let component: HearingListTableComponent;
     const translateService = translateServiceSpy;
     const testData = new ConferenceTestData();
+    let mockedHearingVenueFlagsService: HearingVenueFlagsService;
```
robertbennett1998, line 13: Better to set the type to make it clear its a spy.
```suggestion
    let mockedHearingVenueFlagsService: jasmine.SpyObj<HearingVenueFlagsService>;
```

VideoWeb/VideoWeb/ClientApp/src/app/services/hearing-venue-flags.service.ts
```diff
@@ -0,0 +1,15 @@
+import { Injectable } from '@angular/core';
+import { BehaviorSubject } from 'rxjs';
+
+@Injectable({
+    providedIn: 'root'
+})
+export class HearingVenueFlagsService {
+    private _hearingVenueIsScottish = new BehaviorSubject(false);
```
robertbennett1998, line 8: Do we want this to assume it isn't Scottish wouldn't a ReplaySubject be better?
```suggestion
    private _hearingVenueIsScottish = new ReplaySubject(1);
```

VideoWeb/VideoWeb/ClientApp/src/app/services/hearing-venue-flags.service.ts
```diff
@@ -0,0 +1,15 @@
+import { Injectable } from '@angular/core';
+import { BehaviorSubject } from 'rxjs';
```
robertbennett1998, line 2: If committing the change below:
```suggestion
import { ReplaySubject } from 'rxjs';
```

VideoWeb/VideoWeb/ClientApp/src/app/shared/footer/footer.component.scss
omio-ma: could you double check to make sure thats what you meant, I didnt know actually you could next media queries inside of the class black

robertbennett1998 (changes requested):
A few small suggestions / observations. I agree that this should of been a directive it would massively reduce the overhead of maintaining this particularly from a test point of view. Handling the different locations could be done something like this. 

image

Then if anything needs to change it the maps can be modified. A default crest can be used where no location is provided.

VideoWeb/VideoWeb/ClientApp/src/app/services/hearing-venue-flags.service.ts
omio-ma: I thought it would be safer to assume scottish crest is false by default as we know exactly where and when it should be set to true (ie when a user actually signs into a particular hearing) 

Would it benefit in using replay subject in anyway or is it wrong to use behavior subject here?

FarronF: Why a ReplaySubject? Is the idea that an undefined value would be falsy as well?

I feel like BehaviorSubject makes sense here.

VideoWeb/VideoWeb/ClientApp/src/app/shared/footer/footer.component.scss
omio-ma: > A few small suggestions / observations. I agree that this should of been a directive it would massively reduce the overhead of maintaining this particularly from a test point of view. Handling the different locations could be done something like this.
> 
> image
> 
> Then if anything needs to change it the maps can be modified. A default crest can be used where no location is provided.

Thats the problem with this, not all are actually using images, one of them is a svg and in another place we are using 

> A few small suggestions / observations. I agree that this should of been a directive it would massively reduce the overhead of maintaining this particularly from a test point of view. Handling the different locations could be done something like this.
> 
> image
> 
> Then if anything needs to change it the maps can be modified. A default crest can be used where no location is provided.

few things here, not all are img elements, one of them is a svg, one of the element is using background image inside of a span, some of them are also using different class names. Some components use same crest for scottish venue but different ones for uk etc. things like that has sort of flagged out to me is that it would have ended in being a really complicated directive that would be very hard to maintain.

omio-ma:
> A few small suggestions / observations. I agree that this should of been a directive it would massively reduce the overhead of maintaining this particularly from a test point of view. Handling the different locations could be done something like this.
> 
> image
> 
> Then if anything needs to change it the maps can be modified. A default crest can be used where no location is provided.

few things here, not all are img elements, one of them is a svg, one of the element is using background image inside of a span, some of them are also using different class names. Some components use same crest for scottish venue but different ones for uk etc. things like that has sort of flagged out to me is that it would have ended in being a really complicated directive that would be very hard to maintain.

VideoWeb/VideoWeb/ClientApp/src/app/shared/footer/footer.component.scss
FarronF: It's a feature of SCSS. I believe that wouldn't work with regular CSS

VideoWeb/VideoWeb/ClientApp/src/app/services/hearing-venue-flags.service.ts
robertbennett1998: Anything that relies on it will wait for the value to be emitted. Assuming you aren't blocking calls which I didn't see the image wont load. Then when a value is emitted it will. Rather than the wrong crest been shown first.

robertbennett1998: Replay doesn't emit until it has the first value

FarronF: I think we do want it to show HMCTS crest by default though. Do you think the BehaviorSubject could be an issue say if someone refreshed in the waiting room?

robertbennett1998: Yeah that was what I was thinking as it would show the HMCTS to start with even for a SCTS hearing. Granted  it might not be for long but for a slow connection it could be.

omio-ma: if you hard reload on the waiting room it will initially be hmcts logo and will get changed when we get conference details when getConference gets called which then updates the logo back to scottish crest

omio-ma: you would notice it only if you are really paying attention, however I didnt think that is functionally wrong as by default the app should always show the hmcts crest, and only display the scottish crest based on the hearing venue

FarronF:
Discounting the header and footer as they are handled differently the ones that use src are 

```
hearing-rules.component.ts 
[src]="(hearingVenueIsScottish$ | async) ? '/assets/images/scottish-crest-circle.png': '/assets/images/HmctsCrest.png'"

introduction.component.html 
[src]="(hearingVenueIsInScotland$ | async) ? '/assets/images/scottish-crest-circle.png': '/assets/images/BluePanelIdent.png'"

individual-participant-status-list.component.html 
[src]="(hearingVenueIsInScotland$ | async) ? '/assets/images/scottish-crest-grey.png': '/assets/images/govuk-crest.png'"

judge-participant-status-list.component.html 
[src]="(hearingVenueIsInScotland$ | async) ? '/assets/images/scottish-crest-grey.png': '/assets/images/govuk-crest.png'"

private-consultation-participants.component.html 
[src]="(hearingVenueIsInScotland$ | async) ? '/assets/images/scot-crest-white.png': '/assets/images/UkGovCrestWhite.png'"

participants-panel.component.html 
[src]="(hearingVenueIsInScotland$ | async) ? '/assets/images/scot-crest-white.png': '/assets/images/UkGovCrestWhite.png'"

[src]="(hearingVenueIsInScotland$ | async) ? '/assets/images/scot-crest-white.png': '/assets/images/UkGovCrestWhite.png'"
```

So 4 sets of images. We could store those values in the directive and pass an input to determine which set to use and then inject the service so it's only in one place, would that directive be doing too much then? 

The classes shouldn't be an issue as we can still set those classes the way they are being set.

VideoWeb/VideoWeb/ClientApp/src/app/services/hearing-venue-flags.service.ts
FarronF: It's a very good point. When do we set it though because we need a logo to display in other pages before a user has selected a hearing and stuff that is not required to be authorised for.

FarronF: I think the header/footer displaying a different crest on a reload briefly and whether that is acceptable would need to be  a Neil/Lorenzo question.

FarronF:
Also it looks like HmctsCrest.png and BluePanelIdent.png are the same but different sizes. We could probably remove one of them(confirm with Neil before of course). Which would make 3 sets.

omio-ma:
> Discounting the header and footer as they are handled differently the ones that use src are
> 
> ```
> hearing-rules.component.ts 
> [src]="(hearingVenueIsScottish$ | async) ? '/assets/images/scottish-crest-circle.png': '/assets/images/HmctsCrest.png'"
> 
> introduction.component.html 
> [src]="(hearingVenueIsInScotland$ | async) ? '/assets/images/scottish-crest-circle.png': '/assets/images/BluePanelIdent.png'"
> 
> individual-participant-status-list.component.html 
> [src]="(hearingVenueIsInScotland$ | async) ? '/assets/images/scottish-crest-grey.png': '/assets/images/govuk-crest.png'"
> 
> judge-participant-status-list.component.html 
> [src]="(hearingVenueIsInScotland$ | async) ? '/assets/images/scottish-crest-grey.png': '/assets/images/govuk-crest.png'"
> 
> private-consultation-participants.component.html 
> [src]="(hearingVenueIsInScotland$ | async) ? '/assets/images/scot-crest-white.png': '/assets/images/UkGovCrestWhite.png'"
> 
> participants-panel.component.html 
> [src]="(hearingVenueIsInScotland$ | async) ? '/assets/images/scot-crest-white.png': '/assets/images/UkGovCrestWhite.png'"
> 
> [src]="(hearingVenueIsInScotland$ | async) ? '/assets/images/scot-crest-white.png': '/assets/images/UkGovCrestWhite.png'"
> ```
> 
> So 4 sets of images. We could store those values in the directive and pass an input to determine which set to use and then inject the service so it's only in one place, would that directive be doing too much then?
> 
> The classes shouldn't be an issue as we can still set those classes the way they are being set.

nah that would add complication (if we have it as sets) however one thing that can be done perhaps would be we pass the images as input as well, so we have one attribute that specifies the image source for scottish crest and one that specifies for hmcts

FarronF:
So have inputs for the image sources that are passed into the directive and it decides between the 2 based on the service? That sounds pretty good actually.

VideoWeb/VideoWeb/ClientApp/src/app/services/hearing-venue-flags.service.ts
omio-ma: confirmed with Lorenzo that this isn't a show stopper, a seperate tech debt story has been raised to address this:
https://tools.hmcts.net/jira/browse/VIH-8463

VideoWeb/VideoWeb/ClientApp/src/app/on-the-day/hearing-rules/hearing-rules.component.ts
```diff
@@ -12,6 +13,7 @@ import { Logger } from 'src/app/services/logging/logger-base';
 })
 export class HearingRulesComponent extends ParticipantStatusBaseDirective implements OnInit {
     conferenceId: string;
+    hearingVenueIsScottish$: Observable<boolean>;
```
FarronF, line 16: Can this be removed now?

omio-ma: yea, also spotted few other places I missed out

VideoWeb/VideoWeb/ClientApp/src/app/on-the-day/judge-hearing-list/judge-hearing-list.component.ts
```diff
@@ -52,6 +54,7 @@ export class JudgeHearingListComponent implements OnInit, OnDestroy {
         });
         this.retrieveHearingsForUser();
         this.setupSubscribers();
+        this.hearingVenueFlagsService.setHearingVenueIsScottish(false);
```
FarronF, line 57: Just thinking is there other places a judge could navigate to that should have the default crest? Having it set here might not be good enough. The service might need to subscribe to navigation and check if the route isn't linked to a conference and reset it that way. Or maybe we could look at canDeactivate guards but that doesn't seem like the intent of them but maybe there is something similar.

VideoWeb/VideoWeb/ClientApp/src/app/on-the-day/judge-hearing-list/judge-hearing-list.component.ts
```diff
@@ -98,6 +101,7 @@ export class JudgeHearingListComponent implements OnInit, OnDestroy {
 
     onConferenceSelected(conference: ConferenceForHostResponse) {
         this.logger.debug('[JudgeHearingList] - Signing into judge waiting room', { conference: conference.id });
+        this.hearingVenueFlagsService.setHearingVenueIsScottish(conference.hearing_venue_is_scottish);
```
FarronF, line 104: Similar to above maybe we can have a route guard on the paths related to conferences that would handle this rather than having it in multiple places.

omio-ma: not sure if I am following this, do you mean we have a route guard that we get conference details from?

VideoWeb/VideoWeb/ClientApp/src/app/on-the-day/judge-hearing-list/judge-hearing-list.component.ts
omio-ma: once the judge selects a hearing that is in a scottish venue, we will then render the scottish crest, when they go back to the hearing they will see the hmcts crest again. This is what is in the A/C, anything else we are trying to cover is going to fall out of scope of this ticket, but yea we might want to raise other tickets to address those

VideoWeb/VideoWeb/ClientApp/src/app/on-the-day/judge-hearing-list/judge-hearing-list.component.ts
omio-ma: but as I have mentioned in my comment above, the change I have made is following the A/C and there is possibility that there might be other edge cases we might need to address however it will fall out of the scope of this ticket

FarronF (approved):

VideoWeb/VideoWeb/ClientApp/src/app/on-the-day/judge-hearing-list/judge-hearing-list.component.ts
FarronF: That's fine. I think that could be a more robust solution but given the ACs I'm happy to go ahead with this as is.

FarronF (approved):

VideoWeb/VideoWeb/ClientApp/src/app/on-the-day/hearing-rules/hearing-rules.component.html
```diff
@@ -7,9 +7,13 @@ <h1 class="govuk-heading-xl">{{ 'hearing-rules.court-rules' | translate }}</h1>
...
       <img
-        [src]="(hearingVenueIsScottish$ | async) ? '/assets/images/scottish-crest-circle.png': '/assets/images/HmctsCrest.png'"
+        appCrestLogoImageSource
+        [sctsImageSource]="'/assets/images/scottish-crest-circle.png'"
+        [sctsAltText]=" ('hearing-rules.scottish-crest' | translate)"
+        [hmctsImageSource]="'/assets/images/HmctsCrest.png'"
+        [hmctsAltText]=" ('hearing-rules.hm-courts-and-tribunals-service-crest' | translate)"
         class="govuk-header__logotype-crest crest"
-        [attr.alt]="(hearingVenueIsScottish$ | async) ? ('hearing-rules.scottish-crest' | translate): ('hearing-rules.hm-courts-and-tribunals-service-crest' | translate)"
+        alt=""
```
robertbennett1998, line 16: Why does this need to be here?

robertbennett1998 (approved):
LGTM

VideoWeb/VideoWeb/ClientApp/src/app/on-the-day/hearing-rules/hearing-rules.component.html
omio-ma: sonarcloud flags this as a bug even though we are adding the alt text using javascript

PR #1696, opened 2021-11-04
VIH-8452 disable h264 for iOS users

robertbennett1998:
### JIRA link ###
[VIH-8452](https://tools.hmcts.net/jira/browse/VIH-8452)
[VIH-8445](https://tools.hmcts.net/jira/browse/VIH-8445)

### Change description ###
- Moved code to disable h264 for FireFox to video call service.
- Disabling h264 when using an iOS device.

**Does this PR introduce a breaking change?** 
```
[ ] Yes
[x] No
```

** Notes **
- Also seems to fix the issue in VIH-8445


FarronF (approved):

shaed-parkar:
Have you double-checked the `pexrtc` script doesn't do this already? This was done for firefox but in the script it handles the browsers and sets this false anyway

robertbennett1998:
> Have you double-checked the `pexrtc` script doesn't do this already? This was done for firefox but in the script it handles the browsers and sets this false anyway

The current version doesn't this was a suggested work around by Kinly due to iOS 15.1 issues.

PR #1706, opened 2021-11-10
VIH-8433 Correct default staff member landing page URL

AWildBugAppeared:
https://tools.hmcts.net/jira/browse/VIH-8433

There seems to be a race condition going on on some devices and on Firefox, where the URL isn't being updated correction. I've had a look and the easiest solution seems to be to change the original URL that is set to be the staff member hearing selection instead of the staff member hearing list, which definitely shouldn't be set here anyway.


robertbennett1998 (approved):
LGTM

ZJianWang (approved):

PR #1716, opened 2021-11-22
VIH-8348 Fix hearing telephone participant message in WR

AWildBugAppeared:
https://tools.hmcts.net/jira/browse/VIH-8348

This is a workaround where the root cause is still unidentified. Despite the video element not rendering on the browser, the telephone participant message is still playing.

We have instead updated the logic for muting the video stream so that the stream is muted when the participant is not in the hearing.


VideoWeb/VideoWeb/ClientApp/src/app/shared/force-play-video.directive.ts
```diff
@@ -40,9 +40,9 @@ export class ForcePlayVideoDirective implements OnInit, OnChanges {
 
     private updateMute() {
         if (!!this.mute) {
-            this.renderer.setAttribute(this.videoElement, 'muted', 'true');
+            this.videoElement.muted = true;
         } else {
-            this.renderer.setAttribute(this.videoElement, 'muted', 'false');
+            this.videoElement.muted = false;
```
FarronF, line 45: It would be good if we could use this.renderer.setProperty method instead, we didn't have the chance to test if that would actually work today though.

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/joh-waiting-room/joh-waiting-room.component.html
```diff
@@ -44,7 +44,7 @@
     <video
       appForcePlayVideo
       #incomingFeed
-      (playing)="updateVideoStreamMuteStatus()"
+      [mute]="shouldMuteHearing()"
```
FarronF, line 47: Just to put another option out there, instead of using the directive could we just bind this function to the muted attribute here?

The default behavior is for muted to be false anyway so even the places where we don't pass a value to the mute input should still work as expected.

omio-ma (approved):

ZJianWang (approved):

PR #1717, opened 2021-11-24
VIH-8473 angular fontawesome

FarronF:
VIH-8473 

### Change description ###
A POC of how angular fontawesome package could be used. This package allows tree shaking so we would bundle the icons we define.

```
[ ] Yes
[x] No
```
Current
image

With angular fontawesome, note that the styles applied by the classes works the same.
image


MichaelKeaneathmcts (commented):
Looking at the browser logs for performance, I see that the browser is continually going to fontawesome to download content. why is this not served and cached from HMCTS?

ZJianWang:
I couldn't find any difference in both screenshot , is that right?

FarronF:
@ZJianWang Yeah sorry I should have been more clear with that. Essentially I wanted to show that the custom styles that were applied to those through the classes still worked with this implementation.

@MichaelKeaneathmcts Where do you see this? Essentially we reference an external .js file which is a font awesome kit created by Neil. I'm not entirely clear how this works but it does reference other addresses so I assume when it comes across a font awesome class it will download a css file from there which in turn will reference font files that are also hosted by font awesome that also need to be downloaded. This ticket is regarding issues that have arose around that where certain users are behind a firewall.

Regarding performance this package's advantage is that it supports tree shaking so it will only include icons that we define rather than the full .js file which is over 1MB

ZJianWang:
No worries, Just want double check that, compare with the other silution, looks this one we don't need update our local file along with the libray updating

MichaelKeaneathmcts:
@FarronF You can see this when you run a tool like fiddler, it logs that the fonts are being read from fontawesome.com

PR #1718, opened 2021-11-24
VIH-8473 Font awesome self hosting

FarronF:
VIH-8473

### Change description ###
Putting free font awesome resources in the assets folder. Note this will not work for pro icons.

```
[ ] Yes
[x] No
```


VideoWeb/VideoWeb/ClientApp/src/index.html
```diff
@@ -29,10 +29,11 @@
 
     <meta name="viewport" content="width=device-width, initial-scale=1" />
     <link rel="icon" type="image/x-icon" href="assets/images/favicon.ico" />
+    <link href="assets/fontawesome/css/all.css" rel="stylesheet">
+    <!-- <script defer src="assets/fontawesome/all.js"></script> -->
```
FarronF, line 33: This and all.js could be removed before merging.

FarronF: The js is another option instead of storing css and webfonts. It works with the same classes but instead of importing fonts it draws them as svgs.

PR #1725, opened 2021-11-29
VIH-8526 Judge unable to admit Witness & Witness interpreter into the…

ZJianWang:

VIH-8526 

### Change description ###

if it is interpreter, then callable  return true

```
[ ] Yes
[x] No
```

### DIFF
--- a/VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/models/linked-participant-panel-model.ts
+++ b/VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/models/linked-participant-panel-model.ts
@@ -77,7 +77,7 @@ export class LinkedParticipantPanelModel extends PanelModel {
     }
 
     isAvailable(): boolean {
-        return this.participants.some(p => p.isAvailable());
+        return this.participants.every(p => p.isAvailable());
     }
 
     isInConsultation(): boolean {
--- a/VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/models/linked-participant-panel-model.ts
+++ b/VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/models/linked-participant-panel-model.ts
@@ -46,7 +46,11 @@ export class LinkedParticipantPanelModel extends PanelModel {
     }
 
     get isCallableAndReadyToJoin(): boolean {
-        return this.participants.every(p => p.isCallableAndReadyToJoin);
+        if (this.isWitness) {
+            return this.isAvailable();
+        } else {
+            return this.participants.every(p => p.isCallableAndReadyToJoin);
+        }
     }
 
     get isCallableAndReadyToBeDismissed(): boolean {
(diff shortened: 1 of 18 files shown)


VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/models/participant-panel-model.ts
```diff
@@ -23,7 +23,7 @@ export class ParticipantPanelModel extends IndividualPanelModel {
         return this.isCallable && this.isInHearing();
     }
     get isCallable(): boolean {
-        return this.isWitness || this.isQuickLinkUser;
+        return this.isWitness || this.isQuickLinkUser || this.isInterpreter;
```
FarronF, line 26: I feel like we shouldn't include interpreter as a callable participant as it is dependent on who they are interpreting for. I think we need to instead look at the logic in the linked participant panel model.

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/models/video-endpoint-panel-model.ts
```diff
@@ -2,11 +2,11 @@ import { EndpointStatus, Role, VideoEndpointResponse } from '../../services/clie
 import { IndividualPanelModel } from './individual-panel-model';
 
 export class VideoEndpointPanelModel extends IndividualPanelModel {
-    public status: EndpointStatus;
+    public endpointStatus: EndpointStatus;
```
FarronF, line 5: Is this changed for a reason?

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/models/linked-participant-panel-model.ts
```diff
@@ -77,7 +77,7 @@ export class LinkedParticipantPanelModel extends PanelModel {
     }
 
     isAvailable(): boolean {
-        return this.participants.some(p => p.isAvailable());
+        return this.participants.every(p => p.isAvailable());
```
FarronF, line 80: This may cause an issue with panel members/wingers, Have you tested that?

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/models/participant-panel-model.ts
FarronF: To me the issue looks like it's here in LinkedParticipantPanelModel which is why I understand adding interpreter as callable but I worry that that will behave strangely with non witnesses(I think it might show the dismiss option for them)
```
get isCallableAndReadyToJoin(): boolean {
    return this.participants.every(p => p.isCallableAndReadyToJoin);
}
```

Maybe this should check if any are callable and all are available instead of checking each isCallableAndReadyToJoin.

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/models/video-endpoint-panel-model.ts
ZJianWang: no when I try to see learn the code I got confused with status

FarronF: I don't disagree, I just have a concern because I have seen places where a function will take any type and check ParticipantStatus or EndpointStatus as strings rather than using enums. I would hope the tests would catch something like that but I can't be sure.

kevin-hmcts (approved):
LGTM

FarronF (approved):

PR #1730, opened 2021-12-03
VIH-8524 Calling User API for quick link participants to find the user

ZJianWang:

VIH-8524

### Change description ###

Calling User API for quick link participants to find the user

```
[ ] Yes
[x] No
```

### DIFF
--- a/VideoWeb/VideoWeb.Common/SignalR/AdUserProfileService.cs
+++ b/VideoWeb/VideoWeb.Common/SignalR/AdUserProfileService.cs
@@ -30,10 +30,10 @@ namespace VideoWeb.Common.SignalR
             try
             {
                 var userName = participantUserName;
-                if (!participantUserName.EndsWith(QuickLinkParticipantConst.Domain))
+                if (!IsQuickLinkParticipant(userName))
                 {
                     var profile = await _userApiClient.GetUserByAdUserNameAsync(participantUserName);
-                    userName = profile.FirstName + " " + profile.LastName;
+                    userName = $"{profile.FirstName} {profile.LastName}";
                 }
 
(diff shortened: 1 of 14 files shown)


VideoWeb/VideoWeb.Common/SignalR/AdUserProfileService.cs
```diff
@@ -1,30 +1,41 @@
...
         {
             try
             {
-                var profile = await _userApiClient.GetUserByAdUserNameAsync(username);
-                var userName = profile.FirstName + " " + profile.LastName;
+                var userName = participantUserName;
+                if (!participantUserName.EndsWith(QuickLinkParticipantConst.Domain))
+                {
+                    var profile = await _userApiClient.GetUserByAdUserNameAsync(participantUserName);
+                    userName = profile.FirstName + " " + profile.LastName;
```
omio-ma, line 36: I know this was already implemented like this but can you try using string interpolation here instead: 
```suggestion
                    userName = $"{profile.FirstName} ${profile.LastName}";
```

Make sure you write the test first (or ensure there are tests available) to ensure we are not changing the behavior.

VideoWeb/VideoWeb.Common/SignalR/AdUserProfileService.cs
```diff
@@ -36,9 +47,19 @@ public async Task<string> GetObfuscatedUsernameAsync(string username)
 
         public async Task<UserProfile> GetUserAsync(string username)
         {
+            if (username.EndsWith(QuickLinkParticipantConst.Domain))
```
omio-ma, line 50: create a private function here, thats takes username as an argument and then returns a boolean value, this is because this logic is duplicated here and in line 33

omio-ma (dismissed):

omio-ma (approved):

robertbennett1998 (approved):
LGTM

PR #1739, opened 2021-12-09
VIH-8578 clean up pex rtc when disconnecting from Pexip

robertbennett1998:
[VIH-8578](https://tools.hmcts.net/jira/browse/VIH-8578)

### Change description ###
- As suggested by Alan (from Pexip) we have implemented a method to clean up and recreate the PexRtc instance when the server disconnects, the client disconnects or a (pexip) error is raised. This should hopefully resolve the "Invalid Token" issue. Pexip is not designed to use the same client instance when reconnecting a call the was previously closed.

```
[ ] Yes
[x] No
```

### DIFF
--- a/VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/services/video-call.service.ts
+++ b/VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/services/video-call.service.ts
@@ -86,6 +86,7 @@ export class VideoCallService {
      * the user's preferred camera and microphone (if selected)
      */
     async setupClient(): Promise<void> {
+        this.logger.info(`${this.loggerPrefix} setting up client.`);
         this.hasDisconnected$ = new Subject();
 
         const self = this;
@@ -206,7 +207,7 @@ export class VideoCallService {
     }
 
     private handleError(error: string) {
-        this.kinlyHeartbeatService.stopHeartbeat();
+        this.cleanUpConnection();
 
         this.onErrorSubject.next(new CallError(error));
     }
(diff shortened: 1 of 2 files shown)


VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/services/video-call.service.ts
```diff
@@ -237,14 +235,20 @@ export class VideoCallService {
...
             throw new Error(`${this.loggerPrefix} Pexip Client has not been initialised.`);
         }
     }
 
+    private cleanUpConnection() {
+        this.logger.warn(`${this.loggerPrefix} Cleaning up connection.`);
+        this.hasDisconnected$.next();
+        this.hasDisconnected$.complete();
+        this.kinlyHeartbeatService.stopHeartbeat();
+        this.setupClient();
```
omio-ma, line 249: do we not have access to the instance of the pex rtc and so can we not just set that to null or something rather than setting up client? Or is this done intentionally for some other reason?

kevin-hmcts (approved):
LGTM

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/services/video-call.service.ts
robertbennett1998: We need to re run set up client to set up all the event handlers and the streams.

omio-ma: but wouldn't that get called anyway when its required? So I am wondering isnt there a function that ensures this instance of pexrtc is cleared?

robertbennett1998: No it doesn't it is only done on event hub is ready. This is the cleanest way. No this is what was advised by Alan (from Pexip)

omio-ma (approved):

PR #1749, opened 2021-12-17
VIH-8600 PMs details not displayed in Consultation room participant panel

ZJianWang:

VIH-8600 

### Change description ###

 PMs details not displayed in Consultation room participant panel

```
[ ] Yes
[ x] No
```


VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/waiting-room-shared/wr-participant-list-shared.component.ts
```diff
@@ -205,7 +205,16 @@ export abstract class WRParticipantStatusListDirective implements DoCheck {
...
-        this.panelMembers = this.conference.participants.filter(x => x.hearing_role === HearingRole.PANEL_MEMBER);
+        this.panelMembers = this.conference.participants.filter(x => this.isParticipantPanelMember(x.hearing_role));
+    }
+    protected isParticipantPanelMember(hearingRole: string): boolean {
+        return (
+            hearingRole === HearingRole.MEDICAL_MEMBER ||
+            hearingRole === HearingRole.FINANCIAL_MEMBER ||
+            hearingRole === HearingRole.LEGAL_MEMBER ||
+            hearingRole === HearingRole.DISABILITY_MEMBER ||
+            hearingRole === HearingRole.PANEL_MEMBER
```
FarronF, line 216: Duplication here of what is also in participant-summary.ts isn't ideal. Could cause maintainability issues when adding different roles in the future. Could we get this in a shared function?

FarronF: Maybe we could add the following to hearing-role-model.ts along with the Enum

`export function isPanelMember(hearingRole: string): boolean {
    return (
        hearingRole === HearingRole.MEDICAL_MEMBER ||
        hearingRole === HearingRole.FINANCIAL_MEMBER ||
        hearingRole === HearingRole.LEGAL_MEMBER ||
        hearingRole === HearingRole.DISABILITY_MEMBER ||
        hearingRole === HearingRole.PANEL_MEMBER
    );
}`

And then import this along with HearingRole enum in these 2 places and call it.

Not sure this is "the Angular way" and my I couldn't find an example that matched exactly what we would want but this is loosely based on how the Angular code is setup with a couple of examples of functions used this way in the following files.

https://github.com/angular/angular/blob/master/packages/core/src/sanitization/url_sanitizer.ts
https://github.com/angular/angular/blob/master/packages/compiler/src/chars.ts

PR #1751, opened 2022-01-06
VIH-8535 Reduced bug rate of the staff member status not being correct

AWildBugAppeared:
This reduces the rate of the bug by fixing the property HostInAnotherHearing from returning true in cases when it should be false.


VideoWeb/VideoWeb/Mappings/ParticipantStatusResponseForVhoMapper.cs
```diff
@@ -23,7 +24,8 @@ public class ParticipantStatusResponseForVhoMapper : IMapTo<Conference, IEnumera
                 .Select(x =>
                 {
                     var status = Enum.Parse<ParticipantStatus>(x.ParticipantStatus.ToString());
-                    var hostInHearing = hostsInHearings.Where(j => j.Username == x.Username && j.Id != x.Id);
+                    var hostInHearing = hostsInHearings.Where(j => j.Username == x.Username && j.Id != x.Id && 
+                        (j.Status == ParticipantState.InHearing || j.Status == ParticipantState.Available));
```
FarronF, line 28: Should we check for InConsultation here as well?

AWildBugAppeared: That makes sense. I will add that

ChrisPaulP (approved):

FarronF (approved):

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/participant-waiting-room/participant-waiting-room.component.html
```diff
@@ -6,9 +6,9 @@
   <div class="header-bar">
     <div class="govuk-body" [class]="!isJohRoom && isPrivateConsultation ? 'header-content-grid-with-lock' : 'header-content-grid'">
       <ng-container *ngIf="!isJohRoom && isPrivateConsultation">
-          <div class="room-title">
-              <h1 #roomTitleLabel class="room-title-label">{{ getRoomName() }}</h1>
-          </div>
+        <div class="room-title">
```
omio-ma, line 9: were there no changes for this file could we omit it if thats the case?

VideoWeb/VideoWeb/ClientApp/src/app/vh-officer/vho-shared/participant-status-base/participant-status-base.component.ts
```diff
@@ -101,7 +101,6 @@ export abstract class ParticipantStatusDirective {
         if (!this.participants) {
             return;
         }
```
omio-ma, line 103: same here appears this change was not required?

omio-ma (approved):
leaving this pr approved, I dont see an issue leaving those whitespace changes in.

PR #1759, opened 2022-01-12
VIH-8254 instant messages not appearing in chat window

omio-ma:
### JIRA link ###
https://tools.hmcts.net/jira/browse/VIH-8254 

### Change description ###
- updated event service to update conference ids when a new conference is booked

**Does this PR introduce a breaking change?** 

```
[ ] Yes
[x] No
```

chat window before sending message (for a new conference added while in the command center):
before_im_sent

chat window after sending message (for a new conference added while in the command center):
after_changing_tabs


FarronF:
I may be wrong but I don't feel like we need to push a notification to VHOs and then have them call AddToGroup from the client side. Could we look at a way to do this all on the backend?

omio-ma:
> I may be wrong but I don't feel like we need to push a notification to VHOs and then have them call AddToGroup from the client side. Could we look at a way to do this all on the backend?

Thats what I have tried to do initially, however I have not been able to fetch the connection id in the backend. The connection id is something that comes from the client side so I think it make sense that I didnt find any other way around it.

omio-ma:
I have spent some time researching whether we could fetch the connection id from hub context based on the group name directly before sending a message to the client side, while debugging I couldn't find connection id with the specified client's group and it only appears in the client side when the message is being sent from the handler. The handler is triggered by the video web backend and I couldnt find any way to fetch the connection id other ways.

VideoWeb/VideoWeb.EventHub/Hub/EventHubClient.cs
```diff
@@ -58,12 +58,17 @@ public override async Task OnConnectedAsync()
...
+           var conferenceIds = await GetConferenceIds(isAdmin);
             var tasks = conferenceIds.Select(c => Groups.AddToGroupAsync(Context.ConnectionId, c.ToString())).ToArray();
 
             await Task.WhenAll(tasks);
         }
 
+        public async Task AddToGroup(string conferenceId)
+        {
+            await Groups.AddToGroupAsync(Context.ConnectionId, conferenceId);
+        }
```
FarronF, line 70: I have a slight concern over this being unrestricted. I'm not sure if this is important though. I'm going to take a quick look at if we can do things another way but if we stick with this could we restrict this to VHOs?

omio-ma: I think so there is a function called IsSenderAdmin that might be something I can use, I am not sure i didnt think about that

FarronF:
Me and Anai were looking at this. We also could not see an obvious way to keep this on the backend so we looked a bit more into why there was a group for every conference. And essentially every VHO should be subscribed to every conference ID group, which this achieves. But this seems overly complex and there's no obvious reason why it would be this way and I personally suspect there was a different plan on how to handle subscriptions that was never implemented.

We can talk about this Monday morning. I think the code changes here are totally fine and makes sense with the existing implementation but Anai and I feel we could simplify the code base a good bit by removing this unneeded complexity and changing EventHubClient.SendToAdmin to contain 

```
await Clients.Group(VhOfficersGroupName)
                .ReceiveMessage(dto.Conference.Id, from, dto.To, dto.Message, dto.Timestamp, dto.MessageUuid);
```

kevin-hmcts (approved):
LGTM

PR #1762, opened 2022-01-14
VIH-8631 participant able to join joh consultation

FarronF:
VIH-8631

### Change description ###
Adding check on BFF to ensure user has valid role. Adding guards to UI.

```
[x] Yes
[ ] No
```

Users with role other than judge, judicial office holder or staff member can no longer call consultations/start with a request with VirtualCourtRoomType of JudgeJOH. If they do they will receive a Forbidden response


VideoWeb/VideoWeb.UnitTests/Controllers/ConsultationController/StartConsultationTest.cs
```diff
@@ -236,5 +237,57 @@ public async Task Should_return_exception()
...
+            var controller = GetControllerWithContextForRole(role);
+            var request = ConsultationHelper.GetStartJohConsultationRequest(_testConference);
+            
+            var result =
+                await controller.StartConsultationAsync(
+                    ConsultationHelper.GetStartJohConsultationRequest(_testConference));
+            
+            // Assert
+            int timesCalledExpected;
+            if(shouldBeAllowed)
```
omio-ma, line 262: can you make this two tests

FarronF: I could, but I'd just be duplicating the code before and after this statement. I feel like defining the role and expected result as test cases is a bit more preferable here.

Anaia (approved):

omio-ma (approved):

VideoWeb/VideoWeb.UnitTests/Controllers/ConsultationController/StartConsultationTest.cs
omio-ma: you could have had two tests one that checks when its allowed and one that doesnt to avoid using the logic here. You would still have test cases for the user roles I just meant moving the assertions into two different tests, ie,

` [TestCase(AppRoles.JudicialOfficeHolderRole)]
        [TestCase(AppRoles.JudgeRole)]
        [TestCase(AppRoles.StaffMember)]
        public async Task When_a_user_tries_to_start_judge_joh_consultation_Should_return_correct_response_code_based_on_role(string role)`
        
 `
        [TestCase(AppRoles.CitizenRole)]
        [TestCase(AppRoles.VhOfficerRole)]
        [TestCase(AppRoles.VenueManagementRole)]
        [TestCase(AppRoles.RepresentativeRole)]
        [TestCase(AppRoles.CaseAdminRole)]
        [TestCase(AppRoles.QuickLinkParticipant)]
        [TestCase(AppRoles.QuickLinkObserver)]
        public async Task When_a_user_tries_to_start_judge_joh_consultation_Should_return_correct_response_code_based_on_role(string role)`

omio-ma: but its fine, its just a preference, normally I like avoiding use of if logics in unit tests

PR #1767, opened 2022-01-20
VIH-8618 no camera icon not showing in participant panel

AWildBugAppeared:
https://tools.hmcts.net/jira/browse/VIH-8618


VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/participant-waiting-room/participant-waiting-room.component.ts
```diff
@@ -85,6 +88,12 @@ export class ParticipantWaitingRoomComponent extends WaitingRoomBaseDirective im
 
     ngOnInit() {
         this.init();
+        this.userMediaService.isAudioOnly$.pipe(takeUntil(this.destroyedSubject)).subscribe(async audioOnly => {
+            this.audioOnly = audioOnly;
+
+            const mediaStatus = new ParticipantMediaStatus(false, audioOnly);
```
FarronF, line 94: Can we use false for the is_local_audio_muted parameter here? I would have thought this would cause problems

AWildBugAppeared: I had the same concern but following the logic it seems to work. And I have tested it with the testers as well and it worked fine. 

Essentially, this subscription is only fired in the waiting room when the user toggles on the audioOnly mute on and off, and has no effect on the local mute status inside the hearing as other aspects of that code handle that on hearing start.

FarronF: But isn't the hearing UI also in this component? What I'm concerned with happening is if a user who is audio muted toggles their camera will this set them to unmuted? Or make the UI appear as if they are unmuted?

FarronF (approved):

PR #1784, opened 2022-02-03
VIH-8707 Don't play video if parent element has been destroyed

FarronF:
VIH-8707 

### Change description ###
Adding destroyed property to force play directive that prevents calling play on the element after ngOnDestroy has already been called.

```
[ ] Yes
[x] No
```

### DIFF
--- a/VideoWeb/VideoWeb/ClientApp/src/app/shared/force-play-video.directive.ts
+++ b/VideoWeb/VideoWeb/ClientApp/src/app/shared/force-play-video.directive.ts
@@ -10,6 +10,8 @@ export class ForcePlayVideoDirective implements OnInit, OnDestroy {
     private unsubscribeFromMouseDownCallback: () => void | null = null;
     private unsubscribeFromTouchStartCallback: () => void | null = null;
 
+    private destroyed = false;
+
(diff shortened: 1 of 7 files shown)


VideoWeb/VideoWeb/ClientApp/src/app/shared/force-play-video.directive.ts
```diff
@@ -36,8 +38,12 @@ export class ForcePlayVideoDirective implements OnInit, OnDestroy {
         this.unsubscribeFromTouchStartCallback = this.renderer.listen('window', 'touchstart', this.onMouseDownOrTouchStart.bind(this));
 
         this.videoElement.oncanplay = event => {
-            this.logger.info(`${this.loggerPrefix} - videoElement.oncanplay - playing video`);
-            this.videoElement.play();
+            this.logger.info(`${this.loggerPrefix} - videoElement.oncanplay - event triggered`);
+
+            if (!this.destroyed) {
```
FarronF, line 43: I did wonder if this would still be available after ngOnDestroy was called. We have tested and it works.

kevin-hmcts (approved):
LGTM

PR #1799, opened 2022-02-14
VIH-8745 Endpoint is displayed before witness in JOH consultation

oliver-scott:
https://tools.hmcts.net/jira/browse/VIH-8745

### Change description ###
Fixes an issue where the endpoint participants are displayed before witnesses in the JOH consultation.
Also updates the ordering of the participants in the JOH consultation to match those in the waiting room.
This involves splitting up of witnesses and observers, so that observers can be displayed at the bottom.

```
[ ] Yes
[x] No
```


will-craig (approved):

omio-ma:
Just to confirm one thing here, have we decided to no longer display any participants with the hearing role of "witness" in the consultation room? Or is that being displayed and this is causing a duplication?

oliver-scott:
> Just to confirm one thing here, have we decided to no longer display any participants with the hearing role of "witness" in the consultation room? Or is that already being displayed by some other code and this is causing a duplication?

Hi - witnesses are still being displayed, they are now retrieved as part of getPrivateConsultationParticipants(), which was previously filtering them out and retrieving them as part of getWitnessesAndObservers in order to display them in a different part of the page.

PR #1823, opened 2022-02-25
VIH-8702 hand raised icon not persisting for 2nd host

will-craig:
https://tools.hmcts.net/jira/browse/VIH-8702

### Change description ###
Within the setHandRaiseStatus function - wrapped the update within a promise resolve, on the loadHearingStateForConference function. This is to ensure that the participant first has the latest status from the cache for all participants before sending an update request to the cache and potentially overwriting other participants HandRaised statuses.

Also added async/await to supporting jasmine tests, so that they don't fall through when executing, before promise completes.

```
[ ] Yes
[ x] No
```

### DIFF
--- a/VideoWeb/VideoWeb/ClientApp/src/app/services/conference/video-control-cache.service.ts
+++ b/VideoWeb/VideoWeb/ClientApp/src/app/services/conference/video-control-cache.service.ts
@@ -124,7 +124,7 @@ export class VideoControlCacheService {
                 self.logger.info(`${this.loggerPrefix} re-initialised state for ${this.conferenceId}.`, {
                     hearingControlStates: this.hearingControlStates
                 });
-                raiseHandInCache();
+                setHandRaiseStatusInCache();
             })
             .catch(() => {
                 self.logger.info(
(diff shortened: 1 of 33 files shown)


VideoWeb/VideoWeb/ClientApp/src/app/services/conference/video-control-cache.service.ts
```diff
@@ -114,16 +114,38 @@ export class VideoControlCacheService {
...
+                self.logger.info(
+                    `${this.loggerPrefix} failed to re-initialised state for ${this.conferenceId}.
+                Control States may be out of sync`,
+                    {
+                        hearingControlStates: this.hearingControlStates
+                    }
+                );
+                raiseHandInCache();
+            });
+        const raiseHandInCache = () => {
```
oliver-scott, line 139: Just a thought - if this can raise or lower the hand, maybe name it `setRaiseHandStatusInCache` just so it's clear that it can do both?

oliver-scott (approved):

krishnapottekkat-hmcts (approved):

PR #1847, opened 2022-03-23
VIH-8880 Navigate to error page if device is in use

AWildBugAppeared:
https://tools.hmcts.net/jira/browse/VIH-8880

### DIFF
--- a/VideoWeb/VideoWeb/ClientApp/src/app/services/user-media.service.ts
+++ b/VideoWeb/VideoWeb/ClientApp/src/app/services/user-media.service.ts
@@ -175,18 +175,8 @@ export class UserMediaService {
 
         return this.hasValidCameraAndMicAvailable().pipe(
             take(1),
-            mergeMap(hasValidCameraAndMicAvailable => {
-                if (hasValidCameraAndMicAvailable) {
-                    return this.getCameraAndMicrophoneDevices();
-                } else {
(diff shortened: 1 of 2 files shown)


Anaia (approved):

VideoWeb/VideoWeb/ClientApp/src/app/services/user-media.service.ts
```diff
@@ -199,7 +200,21 @@ export class UserMediaService {
             take(1),
             map(stream => !!stream && stream.getVideoTracks().length > 0 && stream.getAudioTracks().length > 0),
             catchError(error => {
+                console.log(error);
```
will-craig, line 203: this extra console log probably isnt needed as you as you're logging the error already on the line below

will-craig (approved):

will-craig (approved):
Maybe need to add another unit test that mocks a getUserMedia error response, so you can assert the error service is being called when catching it. Will approve, up to you if you want to make those changes and bring the sonarcloud code coverage up from 11%.

oliver-scott (approved):

oliver-scott:
Agree with Will's comments. The code looks ok, if it's possible to add some test coverage that would be great

oliver-scott (approved):

omio-ma (approved):

PR #1861, opened 2022-04-12
VIH-8924 Observer appraiser should not start or join consultations

AWildBugAppeared:
https://tools.hmcts.net/jira/browse/VIH-8924


AWildBugAppeared:
> Just wondering - perhaps this rule should apply to all hearing roles in the Observer case type group (not sure if there are any at the moment), rather than just appraisers?

The ticket was quite specific in what it wanted to exclude which was the observer appraiser. Given there are other options I assume they are supposed to be allowed.

will-craig (dismissed):
lg2m

will-craig (approved):

PR #1869, opened 2022-04-25
VIH-8945 fix for duplicating interpreter

will-craig:
https://tools.hmcts.net/jira/browse/VIH-8934

### Change description ###
When repositioning the interpreter underneath the interpretee, all the indices shifted up one, so upon removing the old interpreter position, the wrong participant would be removed, instead leaving two instances of the interpreter.

```
[ ] Yes
[x ] No
```


Anaia (dismissed):

oliver-scott:
The code looks fine (although might not need the comment - leave that up to you) but could we maybe add some test coverage? It looks like this issue only happens under specific conditions.

oliver-scott (approved):

Anaia (approved):

PR #1890, opened 2022-06-14
VIH-8946 - added trap focus and first element on focus

marcogagliardi:
https://tools.hmcts.net/jira/browse/VIH-8946

Added trap focus and focus to first element


marcogagliardi:
wrong ticket number

PR #1938, opened 2022-12-01
VIH-9508 - sds  prod redirect ur is are being cached

marcogagliardi:
https://tools.hmcts.net/jira/browse/VIH-9508 

added no-cache header to the backend request header

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -290,7 +290,7 @@ resources:
         - repository: templates
           type: github
           name: hmcts/azure-devops-templates
-          ref: 'VIH-9999-Test-stryker-version'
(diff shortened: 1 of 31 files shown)


azure-pipelines.yml
```diff
@@ -27,7 +27,7 @@ parameters:
     - name: RunMutationTests
       displayName: Run Mutation Tests
       type: boolean
-      default: true
+      default: false
```
oliver-scott, line 30: why has this been changed?

marcogagliardi: because we had to disable stryker

oliver-scott (approved):

PR #1959, opened 2023-02-03
VIH-8608 Announce chat box character limit to screen readers

oliver-scott:
https://tools.hmcts.net/jira/browse/VIH-8608

### Change description ###
Adds screen reader accessibility for the waiting room chat box
- Announce character limit for the chat box to screen readers when the chat box has focus
- Announce alert to screen readers when the chat box character limit has been reached or exceeded after a key press, unless the key is delete or backspace. This works by injecting an error message into the div aria region when the alert is to be shown

```
[ ] Yes
[x] No
```


abbasiTahir (approved):

shaed-parkar (approved):
LGTM

PR #1965, opened 2023-02-13
VIH-9307 hide non-video elements

phillipjchambers:
VIH-9307

### Change description ###
Hide all non-video elements when displaying the hearing/consultation room modal/overlay. This is to prevent these elements being in the DOM and therefore visible to screen-readers, reducing the accessibility of the application.

```
[ ] Yes
[x] No
```


shaed-parkar (approved):
Nice work

PR #1989, opened 2023-03-28
VIH-9828 Adjust chat message font colour for light backgrounds

oliver-scott:
https://tools.hmcts.net/jira/browse/VIH-9828

### Change description ###
Fixes an issue where the font for the chat message, when viewed from the waiting room, is white on a white background and so is not visible.
Applies a condition so that we can use a white front when in the hearing, and the default black font when in the waiting room.

```
[ ] Yes
[x] No
```


marcogagliardi (approved):
👌🏽

PR #1994, opened 2023-04-04
VIH-9562 Enable login with DOM1

shaed-parkar:

VIH-9562

### Change description ###

* Allow users to log in with DOM1 accounts
* Map user role in Bookings API to the app role and add the role claim

```
[ ] Yes
[x] No
```


shaed-parkar:
replaced by #2012

PR #2013, opened 2023-05-12
VIH-9835 mapped wrong object type

will-craig:


shaed-parkar (changes requested):
The docs state this is an object with a single field UUID. A class wit a single field `uuid` would align with docs and also better protect the app from future changes in case pexip change the returned object in anyway.
https://docs.pexip.com/api_client/api_pexrtc.htm#onParticipantDelete

VideoWeb/VideoWeb/ClientApp/src/typings.d.ts
```diff
@@ -39,7 +39,7 @@ declare interface PexipClient {
     onDisconnect: (reason: any) => void;
     onParticipantCreate: (participantUpdate: PexipParticipant) => void;
     onParticipantUpdate: (participantUpdate: PexipParticipant) => void;
-    onParticipantDelete: (participantUpdate: PexipParticipant) => void;
+    onParticipantDelete: (uuid: string) => void;
```
shaed-parkar, line 42: To be consistent here, may I suggest `PexipParticipantDeleted`
```suggestion
    onParticipantDelete: (participantUpdate: PexipParticipantDeleted) => void;
```

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/services/video-call.service.ts
```diff
@@ -44,7 +44,7 @@ export class VideoCallService {
     private onParticipantUpdatedSubject = new Subject<ParticipantUpdated>();
     private onConferenceUpdatedSubject = new Subject<ConferenceUpdated>();
     private onParticipantCreatedSubject = new Subject<ParticipantUpdated>();
-    private onParticipantDeletedSubject = new Subject<ParticipantUpdated>();
+    private onParticipantDeletedSubject = new Subject<string>();
```
shaed-parkar, line 47: To keep in line with what the updated and created does, use a DTO

```suggestion
    private onParticipantDeletedSubject = new Subject<ParticipantDeleted>();
```

shaed-parkar (approved):

komaldabhi1509 (approved):

PR #2016, opened 2023-05-24
VIH-9923 subscribe to participant list change in command centre

shaed-parkar:

VIH-9923

### Change description ###

update participant list for selected conference in the command centre

```
[ ] Yes
[x] No
```


VideoWeb/VideoWeb/ClientApp/src/app/shared/models/hearing.ts
```diff
@@ -10,8 +10,10 @@ import { HearingBase } from './hearing-base';
 import { Participant } from './participant';
 
 export class Hearing extends HearingBase {
+    _participants: Participant[];
```
will-craig, line 13: I think this array should probably be private as it looks like it's only used within the scope of the component and not in the view. Also not sure how I feel about _name convention can it be given a name that diffrentiates it from the other participants array?

will-craig (approved):

will-craig (approved):

PR #2023, opened 2023-06-08
VIH-9441 Fix for quick links error

oliver-scott:
https://tools.hmcts.net/jira/browse/VIH-9941

### Change description ###
Fixes an error when using the quick links url to join a hearing.
This reverts some changes made as part of the angular 15 upgrade. By using `combineLatest` with the idp and security provider, it can return an invalid combination (quick link config + oidc service) as they can change independently, causing the error. To remedy this, revert back to the original mechanism where we get the client settings, and then the idp and security provider together.
Also removes an unused method (`checkAuth`) and its test.

```
[ ] Yes
[x] No
```


shaed-parkar (approved):
Nice find!

komaldabhi1509 (approved):

PR #2082, opened 2023-10-16
VIH-10249 Bug fix allocation notification when a judge is not yet assigned

shaed-parkar:
### Jira link (if applicable)

VIH-10249

### Change description ###

Ignore the judge name when the judge value is null. 

----------------
Before:
image

After:
image


will-craig (approved):

shaed-parkar:
> it's good that we're handling a null judge (which I guess we'll need for ejud) but do we know why the judge is null in the example shown in the ticket? Just wondering if maybe the judge name isn't being passed in

I thought the was the fix until I realised ejud is not on yet. I've ran all the tests and the judge display name is being passed through. annoyingly I cannot debug this due to devops restrictions that still haven't been resolved yet. could we pair on this?

will-craig (approved):

PR #2094, opened 2023-10-26
VIH-7387 Host chooses to enter hearing muted

oliver-scott:
### Jira link (if applicable)
https://tools.hmcts.net/jira/browse/VIH-7387

### Change description ###

- Add a checkbox to the start hearing/resume hearing form for a host to start the hearing muted. This setting is preserved in local storage. The checkbox is only available when the `host-mute-microphone` feature is enabled
- Display the same confirmation when joining a hearing
- Conference settings associated with conferences for the previous day are removed upon loading the form, to prevent taking up unnecessary space in local storage


VideoWeb/VideoWeb/ClientApp/src/app/services/conference/local-storage.service.ts
```diff
@@ -8,7 +8,7 @@ export class LocalStorageService {
     private loggerPrefix = '[LocalStorageService] -';
     constructor(private logger: LoggerService) {}
 
-    save<T extends object>(key: string, value: T, overwrite: boolean = true): boolean {
+    save(key: string, value: string | object, overwrite: boolean = true): boolean {
```
will-craig, line 11: Just trying to understand the logic here; is this change because you want to store a string a value? But because string is a primitive type in js, its not a type of obect?

will-craig (dismissed):
lg2m :)

will-craig (approved):

PR #2095, opened 2023-11-08
VIH-10308 Bugfix unable to send IM as Staff Member once joined to a conference

shaed-parkar:
### Jira link (if applicable)

VIH-10308

### Change description ###

When a SM with the VHO role joins a hearing, they will pass the condition `IsParticipantInConference`. This rule should be updated to check the role is not staff member so that the user can still be treated with their VHO role


oliver-scott (approved):

himesh99 (approved):
lgtm

PR #2105, opened 2023-12-18
VIH-10376 New users seeing you do not have a hearing page

will-craig:
### Jira link (if applicable)
https://tools.hmcts.net/jira/browse/VIH-10376

### Change description ###
Changes get conferences for individual and host endpoints, to get the hearing ids from booking api first then query the conferences with the hearing ids, rather than the username

### DIFF
--- a/VideoWeb/VideoWeb.Common/VideoWeb.Common.csproj
+++ b/VideoWeb/VideoWeb.Common/VideoWeb.Common.csproj
@@ -20,7 +20,7 @@
     </PackageReference>
     <PackageReference Include="StackExchange.Redis" Version="2.6.111" />
     <PackageReference Include="System.IdentityModel.Tokens.Jwt" Version="5.6.0" />
-    <PackageReference Include="VideoApi.Client" Version="1.48.7" />
+    <PackageReference Include="VideoApi.Client" Version="1.50.1-pr-0626-0005" />
   </ItemGroup>
     
 </Project>
--- a/VideoWeb/VideoWeb/Mappings/BookingForHostResponseMapper.cs
+++ b/VideoWeb/VideoWeb/Mappings/BookingForHostResponseMapper.cs
@@ -20,23 +20,20 @@ public class BookingForHostResponseMapper : IMapTo<ConfirmedHearingsTodayRespons
     public Contract.Responses.ConferenceForHostResponse Map(ConfirmedHearingsTodayResponse booking, List<ConferenceForHostResponse> conferences)
     {
         var conference = conferences.Find(x => x.HearingId == booking.Id);
-        var dto = new Contract.Responses.ConferenceForHostResponse
-        {
-            Id = conference.Id,
-
-            CaseName = booking.CaseName,
-            CaseNumber = booking.CaseNumber,
-            CaseType = booking.CaseTypeName,
-            ScheduledDuration = booking.ScheduledDuration,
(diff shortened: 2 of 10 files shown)


VideoWeb/VideoWeb/Mappings/BookingForHostResponseMapper.cs
```diff
@@ -20,23 +20,20 @@ public BookingForHostResponseMapper(IMapTo<ParticipantForHostResponse, VideoWeb.
...
-            ScheduledDateTime = booking.ScheduledDateTime,
-            NumberOfEndpoints = booking.Endpoints.Count,
-            HearingVenueIsScottish = booking.IsHearingVenueScottish,
-            Status = Enum.Parse<ConferenceStatus>(conference.Status.ToString()),
-            ClosedDateTime = conference.ClosedDateTime,
-            Participants =
-                conference.Participants.Select(_participantForHostResponseMapper.Map)
-                    .ToList() // need to use participant list from video api to include QL users
-        };
+        var dto = new Contract.Responses.ConferenceForHostResponse();
```
oliver-scott, line 23: are these changes needed? Probably best to use the object initialiser if we can

charts/vh-video-web/values.dev.template.yaml
```diff
@@ -10,4 +10,5 @@ java:
         EJUDAD__POSTLOGOUTREDIRECTURI: https://${SERVICE_FQDN}/logout
         EJUDAD__REDIRECTURI: https://${SERVICE_FQDN}/home
         DOM1__POSTLOGOUTREDIRECTURI: https://${SERVICE_FQDN}/logout
-        DOM1__REDIRECTURI: https://${SERVICE_FQDN}/home
\ No newline at end of file
+        DOM1__REDIRECTURI: https://${SERVICE_FQDN}/home        
+        VHSERVICES__VIDEOAPIURL: https://vh-video-api-pr-626.dev.platform.hmcts.net/
```
oliver-scott, line 14: just a reminder to change before the merge

VideoWeb/VideoWeb.Common/VideoWeb.Common.csproj
```diff
@@ -20,7 +20,7 @@
     </PackageReference>
     <PackageReference Include="StackExchange.Redis" Version="2.6.111" />
     <PackageReference Include="System.IdentityModel.Tokens.Jwt" Version="5.6.0" />
-    <PackageReference Include="VideoApi.Client" Version="1.48.7" />
+    <PackageReference Include="VideoApi.Client" Version="1.50.1-pr-0626-0005" />
```
oliver-scott, line 23: just a reminder to change before the merge

oliver-scott (approved):
nicely done. Just a minor observation added, not a dealbreaker though

VideoWeb/VideoWeb/Mappings/BookingForHostResponseMapper.cs
will-craig: I actually prefer it this way, because if it throws an exception you can see immediately which line it was on, and which property has the null ref, rather than it just pointing to line 23, and they you have to work it out. Would save debug time if we didnt use object initialisers everywhere

oliver-scott: understandable

PR #2113, opened 2024-02-14
VIH-10491 Removed wowza polling and updated logic for validating audio streaming

will-craig:
### Jira link (if applicable)
https://tools.hmcts.net/jira/browse/VIH-10491

### Change description ###
Have removed the polling mechanism that sends an api request to validate the recording stream. We now rely on a property on the wowza streaming agent 'isAudioCallOnly' will be set to true if connection established correctly.


marcogagliardi (approved):
lg2me

komaldabhi1509 (approved):

PR #2114, opened 2024-02-14
VIH-10475 A new participant being added resets judges custom display name

will-craig:
### Jira link (if applicable)
https://tools.hmcts.net/jira/browse/VIH-10475

### Change description ###
Updated displayname update to also send a participantUpdated event to other participants to show new display name


VideoWeb/VideoWeb/Controllers/ParticipantsController.cs
```diff
@@ -137,14 +137,13 @@ public async Task<IActionResult> GetHeartbeatDataForParticipantAsync(Guid confer
...
-            [FromBody] UpdateParticipantDisplayNameRequest participantRequest)
+        public async Task<IActionResult> UpdateParticipantDisplayNameAsync(Guid conferenceId, Guid participantId, [FromBody] UpdateParticipantDisplayNameRequest participantRequest)
         {
             try
             {
-                var apiRequest = _mapperFactory.Get<UpdateParticipantDisplayNameRequest, UpdateParticipantRequest>()
-                    .Map(participantRequest);
-                await _videoApiClient.UpdateParticipantDetailsAsync(conferenceId, participantId, apiRequest);
+                var apiRequest = _mapperFactory.Get<UpdateParticipantDisplayNameRequest, UpdateParticipantRequest>().Map(participantRequest);
+                _videoApiClient.UpdateParticipantDetailsAsync(conferenceId, participantId, apiRequest).Wait(); //wait for update to complete before fetching participants to notify
```
oliver-scott, line 145: can we not use `await` here instead?

`await _videoApiClient.UpdateParticipantDetailsAsync(conferenceId, participantId, apiRequest);`

VideoWeb/VideoWeb/Controllers/ParticipantsController.cs
```diff
@@ -165,6 +179,7 @@ public async Task<IActionResult> GetHeartbeatDataForParticipantAsync(Guid confer
         [HttpGet("{conferenceId}/vhofficer/participants")]
         [ProducesResponseType(typeof(IEnumerable<ParticipantContactDetailsResponseVho>), (int)HttpStatusCode.OK)]
         [ProducesResponseType((int)HttpStatusCode.BadRequest)]
+        [ProducesResponseType((int)HttpStatusCode.BadRequest)]
```
oliver-scott, line 182: not sure a second BadRequest annotation is needed here

VideoWeb/VideoWeb.UnitTests/Controllers/ParticipantController/UpdateJudgeDisplayNameTests.cs
```diff
@@ -13,9 +14,14 @@
...
 using EventHubEventType = VideoWeb.EventHub.Enums.EventType;
 using Autofac.Extras.Moq;
+using VideoApi.Contract.Enums;
 using VideoWeb.Contract.Request;
 using VideoWeb.Mappings;
 using VideoApi.Contract.Requests;
+using VideoApi.Contract.Responses;
+using VideoWeb.Contract.Responses;
+using VideoWeb.EventHub.Models;
+using LinkedParticipantResponse = VideoWeb.Contract.Responses.LinkedParticipantResponse;
```
oliver-scott, line 24: can we remove this, doesn't look like it is used

`using LinkedParticipantResponse = VideoWeb.Contract.Responses.LinkedParticipantResponse;`

oliver-scott (commented):
looks good, just a few comments

VideoWeb/VideoWeb/Controllers/ParticipantsController.cs
will-craig: I used wait, as the next statement fetches the updated conference data, and wanted to be sure this request completes before call ing the get request to retrieve the updated conference data

will-craig: Actually, I think you're right, shouldn't make a difference actually if it awaits the task completion before then next await. Was thinking about it in the wrong way

oliver-scott: but I think await should wait for the completion too. Just a bit conscious that `Wait()` could lead to deadlocks as it blocks the thread.

does await not work?

oliver-scott: sorry sent that last message at the last time as yours, disregard :)

oliver-scott (approved):

komaldabhi1509 (approved):

PR #2129, opened 2024-04-16
VIH-10623 Add missing case type translations

oliver-scott:
### Jira link (if applicable)
https://tools.hmcts.net/jira/browse/VIH-10623

### Change description ###
Adds some case type translations that were missing


shaed-parkar (approved):
Do we have a list of all case types? Or maybe create a script to pull all values from Bookings API and check for translations. This issue pops up quite often

komaldabhi1509 (approved):

oliver-scott:
> Do we have a list of all case types? Or maybe create a script to pull all values from Bookings API and check for translations. This issue pops up quite often

yes that's a good shout, maybe we can have a separate ticket to check all case types. Hopefully this will cover at least a good chunk of them, as this issue arose from a specific change where case types were renamed

PR #2154, opened 2024-05-31
VIH-10723 get end of day active sessions

shaed-parkar:
### Jira link (if applicable)

VIH-10723

### Change description ###

* Treat VH Team leads as Administrators (new app role to video web)
* Display a button to show active sessions when user is an Administrator
* The claims cache was unable to deserialise because the Claims does not have a default ctor, resulting in additional API calls

image


charts/vh-video-web/values.dev.template.yaml
```diff
@@ -10,4 +10,5 @@ java:
         EJUDAD__POSTLOGOUTREDIRECTURI: https://${SERVICE_FQDN}/logout
         EJUDAD__REDIRECTURI: https://${SERVICE_FQDN}/home
         DOM1__POSTLOGOUTREDIRECTURI: https://${SERVICE_FQDN}/logout
-        DOM1__REDIRECTURI: https://${SERVICE_FQDN}/home
\ No newline at end of file
+        DOM1__REDIRECTURI: https://${SERVICE_FQDN}/home
```
shaed-parkar, line 13: reminder to revert

VideoWeb/VideoWeb/Mappings/ClaimsPrincipalToUserProfileResponseMapper.cs
```diff
@@ -17,36 +17,33 @@ public UserProfileResponse Map(ClaimsPrincipal user)
...
         return response;
     }
-
+    
     private static List<Role> DetermineRolesFromClaims(ClaimsPrincipal user)
     {
-        var roles = new List<Role>();
+        var roleChecks = new Dictionary<Role, Func<ClaimsPrincipal, bool>>
+        {
+            { Role.VideoHearingsOfficer, u => u.IsInRole(AppRoles.VhOfficerRole) },
```
shaed-parkar, line 30: Sonar complained about the rising cognitive complexity with an additional is statement, so I refactored this into a dictionary

VideoWeb/VideoWeb.Common/Models/CustomClaim.cs
```diff
@@ -0,0 +1,17 @@
+namespace VideoWeb.Common.Models;
+
+/// <summary>
+/// This class is to represent a parameterless constructor for the Claim class.
+/// </summary>
+public class CustomClaim
```
shaed-parkar, line 6: Added this to resolve:
`Newtonsoft.Json.JsonSerializationException: Unable to find a constructor to use for type System.Security.Claims.Claim. A class should either have a default constructor, one constructor with arguments or a constructor marked with the JsonConstructor attribute`.

Suspected we couldn't deserialise and so called the API again anyway

oliver-scott (commented):
left some comments about the roles, looks good otherwise. I'm sure this will be a useful feature

VideoWeb/VideoWeb.Common/AppRoleService.cs
```diff
@@ -95,6 +94,11 @@ private static List<Claim> MapUserRoleToAppRole(List<JusticeUserRole> userRoles)
                 }
             }
             
+            if (userRoles.Exists(x => x == JusticeUserRole.VhTeamLead))
+            {
+                // Team leaders (Admins) are also VHOs so add the VHO role to get the same journey
+                claims.Add(new Claim(ClaimTypes.Role, AppRoles.VhOfficerRole));
```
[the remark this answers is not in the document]
shaed-parkar: Worked this out with Lucy in a call. When a user only has Admin and nothing else they get 401 after log in because they don't fit any of the roles in the `navigator.component`. Currently only admin web cares about the difference between Admin and CSO. In Video web we mapped both roles to VHO. What we want for VHO is an admin to have the same journey as a CSO with a few extra privileges. It's easier to add the CSO role to provide the same journey than to go through the code and add `if role.contains(Admin)` in existing conditions. Let's have a call to go over it anyway.

oliver-scott (approved):

PR #2160, opened 2024-06-11
VIH-10687 Add visual cue for IOS users

oliver-scott:
### Jira link (if applicable)
https://tools.hmcts.net/jira/browse/VIH-10687

### Change description ###

- Display a confirmation popup to iPhone and iPad users (participants and johs) when joining the waiting room
- Move `yes-no-popup-base` component into shared folder
- Remove safari browser check from `isIpad()`


VideoWeb/VideoWeb/ClientApp/src/app/services/device-detection.service.ts
```diff
@@ -0,0 +1,26 @@
+import { Injectable } from '@angular/core';
+import { Logger } from 'src/app/services/logging/logger-base';
+
+@Injectable({
+    providedIn: 'root'
+})
+export class DeviceDetectionService {
```
shaed-parkar, line 7: there should already be a device detector service (think it's called `DeviceTypeService`) that wraps the device type library. might want to check that out

shaed-parkar (commented):
Looks good but I think you may have reinvented the wheel here a bit

VideoWeb/VideoWeb/ClientApp/src/app/services/device-detection.service.ts
oliver-scott: thanks. Have changed the implementation to use that service instead, and removed the new one

shaed-parkar (approved):
Looks good, let's hope that sorts this long term issue out!

komaldabhi1509 (approved):

SwapnaVersion1 (approved):

PR #2171, opened 2024-06-20
VIH-10215 Conference Service and participant details refactor

will-craig:
### Jira link (if applicable)
https://tools.hmcts.net/jira/browse/VIH-10215

### Change description ###
Introduces a conference service to funnel requests through; ensuring conference data is retrieved in a singular way to reduce complexity of using two database sources, and remove the number of dependencies across the application. Service also introduces a 'ForceGetConference' method, to ensure the service restores the latest to the cache, before fetching. Cache dto now maps responses from both bookings-api and video-api. 

Refactored the CourtRoomMapper: Due to not having access to the judge first/last name for building this response, needed to fully rewrite this; seemed to be entirely wrong implementation anyway, as it should be sorting by venue then judge, not judge first name then last name.

Removed CaseTypeGroup entirely. CaseType is a legacy concept that should've been removed with the booking ref data changes. Most participants is simply set to None anyway, so updated any logic that relied on this to use hearing or user role values instead.

Design:
conferenceService drawio
conferenceService drawio


VideoWeb/VideoWeb.Common/VideoWeb.Common.csproj
```diff
@@ -16,7 +16,7 @@
     </PackageReference>
     <PackageReference Include="StackExchange.Redis" Version="2.7.33" />
     <PackageReference Include="System.IdentityModel.Tokens.Jwt" Version="7.5.2" />
-    <PackageReference Include="VideoApi.Client" Version="1.51.10" />
+    <PackageReference Include="VideoApi.Client" Version="2.1.3-pr-0651-0011" />
```
oliver-scott, line 19: just a reminder to update before merge

oliver-scott (approved):

ronaldmalone (approved):

PR #2173, opened 2024-06-21
VIH-10754 Fix for participant room getting lost on updates

oliver-scott:
### Jira link (if applicable)
https://tools.hmcts.net/jira/browse/VIH-10754

### Change description ###
Fixes an issue where the current room for participants in a consultation is lost when new participants are added to the booking. Save the room to the conference cache when a participant/endpoint is transferred into or out of it

Also fixes an infinite loop for authentication where video web continuously tries and fails to GetJusticeUser for non-justice users


will-craig (approved):
Looks good, going through it looks we've done alot of the same stuff in the conference dto - I prefer the way you've done your CurrentRoom class with the room methods, so will keep your implementation when merging this

will-craig (approved):

komaldabhi1509 (approved):

PR #2175, opened 2024-06-24
VIH-10779 observers skip self test and connect to receive stream only

shaed-parkar:
### Jira link (if applicable)

VIH-10779 

### Change description ###

* Connect to pexip use with call_type 'recvonly' when participant is an observer
* Observers skip the self test


oliver-scott (approved):

shaed-parkar:
journey changed. replaced by #2184

komaldabhi1509 (commented):
Approved

PR #2182, opened 2024-06-26
VIH-10668 clerks connect without devices and then upgrades and downgrades

shaed-parkar:
### Jira link (if applicable)

VIH-10668

### Change description ###

* clerks connect without an outgoing stream
* Clerks can then upgrade their call to a normal video call
* Clerks can then downgrade their call to a receive only


shaed-parkar:
journey changed. replaced by #2184

PR #2184, opened 2024-06-26
VIH-10668 clerk and observers join journey

shaed-parkar:
### Jira link (if applicable)

VIH-10668

### Change description ###

* Users will be able to connect to Pexip without a camera and microphone and join as receiveonly
* Users will connect as audio only with no camera is available
* Users will connect as usual when camera and microphone are available
* Observers will skip self test and join a hearing as receive-only
* Change devices/Toggle devices will show or hide based on the presence of a camera or microphone


will-craig (approved):
Looks good, nice one :)

komaldabhi1509 (approved):

PR #2207, opened 2024-08-08
VIH-10833 Fix for quick link users showing as disconnected when joining after the judge

oliver-scott:
### Jira link (if applicable)
https://tools.hmcts.net/jira/browse/VIH-10833

### Change description ###
Reverts a recent change to fetching participants - use `ForceGetConference` to get them from the db instead of the cache when building the participants list, to ensure participants have the latest statuses in the hearing

### DIFF
--- a/VideoWeb/VideoWeb/Controllers/ParticipantsController.cs
+++ b/VideoWeb/VideoWeb/Controllers/ParticipantsController.cs
@@ -75,9 +75,6 @@ namespace VideoWeb.Controllers
             try
             {
                 await videoApiClient.RaiseVideoEventAsync(conferenceEventRequest);
-                
-                // Refresh the conference in the cache
-                await conferenceService.ForceGetConference(conferenceId);
 
                 return NoContent();
             }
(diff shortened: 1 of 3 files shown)


VideoWeb/VideoWeb/Controllers/ParticipantsController.cs
```diff
@@ -175,7 +175,7 @@ public async Task<IActionResult> GetParticipantsWithContactDetailsByConferenceId
             }
             try
             {
-                var conference = await conferenceService.GetConference(conferenceId);
+                var conference = await conferenceService.ForceGetConference(conferenceId);
```
will-craig, line 178: This endpoint is the one used in the command centre by the vhofficer? Does this need to be updated aswell

oliver-scott: good point, thanks. Have reverted the change

komaldabhi1509 (approved):

PR #2213, opened 2024-08-19
VIH-10898 bugfixes post package upgrades

shaed-parkar:
### Jira link

VIH-10898 bugfix import for heartbeat

### Change description

update import for heartbeat js directly rather via angular
add global typings for pexrtc and heartbeat js to simplify component imports
init accordions as advised by the post v5 upgrade for GDS


marcogagliardi (approved):
LG2M

ronaldmalone (approved):

PR #2214, opened 2024-08-20
VIH-10863 Use conference supplier

oliver-scott:
### Jira link
https://tools.hmcts.net/jira/browse/VIH-10863

### Change description
Follow on from [VIH-10807](https://tools.hmcts.net/jira/browse/VIH-10807). Use the supplier from the conference instead of the feature toggle


VideoWeb/VideoWeb.UnitTests/VideoWeb.UnitTests.csproj
```diff
@@ -22,7 +22,7 @@
     <PackageReference Include="NUnit" Version="4.1.0" />
     <PackageReference Include="NUnit3TestAdapter" Version="4.5.0" />
     <PackageReference Include="NBuilder" Version="6.1.0" />
-    <PackageReference Include="VideoApi.Client" Version="2.2.1" />
+    <PackageReference Include="VideoApi.Client" Version="2.2.6-pr-0667-0002" />
```
oliver-scott, line 25: TODO update before merge

charts/vh-video-web/values.dev.template.yaml
```diff
@@ -11,3 +11,4 @@ java:
         EJUDAD__REDIRECTURI: https://${SERVICE_FQDN}/home
         DOM1__POSTLOGOUTREDIRECTURI: https://${SERVICE_FQDN}/logout
         DOM1__REDIRECTURI: https://${SERVICE_FQDN}/home
+        VHSERVICES__VIDEOAPIURL : https://vh-video-api-pr-667.dev.platform.hmcts.net/
```
oliver-scott, line 14: revert before merge

VideoWeb/VideoWeb.Common/SupplierPlatformServiceFactory.cs
```diff
@@ -0,0 +1,43 @@
...
+            
+            return new SupplierPlatformService(jwtTokenProvider, supplierConfiguration);
+        }
+
+        private IJwtTokenProvider GetJwtTokenProvider(Supplier supplier) =>
+            supplier switch
+            {
+                Supplier.Kinly => serviceProvider.GetService<IKinlyJwtTokenProvider>(),
+                Supplier.Vodafone => serviceProvider.GetService<IVodafoneJwtTokenProvider>(),
+                _ => throw new InvalidOperationException($"Unsupported supplier {supplier}")
```
shaed-parkar, line 32: since it's an enum I don't think we're going to be able to hit this line. probably better to merge kinly and default to the Kinly implementation

VideoWeb/VideoWeb.Contract/Responses/ClientSettingsResponse.cs
```diff
@@ -14,25 +15,6 @@ public class ClientSettingsResponse
         /// The eventhub path
         /// </summary>
         public string EventHubPath { get; set; }
-        /// <summary>
```
shaed-parkar, line 17: Good call, I noticed we don't need or use this in video web when setting up the docker images

VideoWeb/VideoWeb.UnitTests/Common/SupplierPlatformServiceFactoryTests.cs
```diff
@@ -0,0 +1,64 @@
...
+
+            // Act
+            var service = sut.Create(supplier);
+
+            // Assert
+            service.Should().BeOfType<SupplierPlatformService>();
+            service.Should().NotBeNull();
+            var tokenProvider = service.GetTokenProvider();
+            var supplierConfiguration = service.GetSupplierConfiguration();
+            switch (supplier)
```
shaed-parkar, line 51: This test should be split up into two. a switch statement in assertions might become a headache to maintain later. That or make the types part of the test case parameters

VideoWeb/VideoWeb.UnitTests/Controllers/ConfigSettingController/GetClientConfigurationSettingsTests.cs
```diff
@@ -65,13 +72,31 @@ public void Should_return_response_with_settings()
...
+        var vodafoneConfiguration = new VodafoneConfiguration
+        {
+            JoinByPhoneFromDate = "2022-02-09"
+        };
+
+        var supplierConfigurations = new List<SupplierConfiguration>
+        {
+            kinlyConfiguration
+        };
+        if (vodafoneEnabled)
```
shaed-parkar, line 84: if statements in setup makes for a confusing setup. I thin we should have a test for true and false with a single flow rather than conditional ones

shaed-parkar (commented):
Great change, minimising our dependency on the toggle. Just a few comments to checkout

VideoWeb/VideoWeb.UnitTests/Controllers/ConfigSettingController/GetClientConfigurationSettingsTests.cs
oliver-scott: split into separate tests

VideoWeb/VideoWeb.Common/SupplierPlatformServiceFactory.cs
shaed-parkar: This is probably a tricky one to balance but I agree this will protect us from any future enum values being added but not being properly implemented through

shaed-parkar (approved):

ronaldmalone (approved):

PR #2217, opened 2024-08-23
VIH-10909 bug fix - events not publishing the latest status

shaed-parkar:
### Jira link

VIH-10909

### Change description

* Update cache before publishing the new statuses
* Update assertions to
* Introduce a globalusings for the EventHub project to reduce using statements


VideoWeb/VideoWeb.EventHub/Handlers/TransferEventHandler.cs
```diff
@@ -25,32 +19,32 @@ public class TransferEventHandler(
 
         protected override async Task PublishStatusAsync(CallbackEvent callbackEvent)
         {
-            var participantStatus = DeriveParticipantStatusForTransferEvent(callbackEvent);
+            var participantStatusTuple = DeriveParticipantStatusForTransferEvent(callbackEvent);
```
oliver-scott, line 22: could we keep this as `participantStatus`?

oliver-scott (approved):
one small suggested comment but not a deal breaker. Looks good

ronaldmalone (approved):

VideoWeb/VideoWeb.EventHub/Handlers/TransferEventHandler.cs
shaed-parkar: I renamed it to tuple so it's clear what we're getting back. Ideally I want to remove the duplicate models when they're both the same

PR #2219, opened 2024-08-27
VIH-10839 Replace polling for hearings with events

oliver-scott:
### Jira link
https://tools.hmcts.net/jira/browse/VIH-10839

### Change description
Replaces polling for hearings on the hearing list pages with subscriptions to events in order to automatically refresh the list.

Messages are now also pushed to the staff members group to facilitate the auto-refresh on the staff members hearing list page.


VideoWeb/VideoWeb/Controllers/InternalEventController.cs
```diff
@@ -171,5 +177,25 @@ public async Task<IActionResult> AllocationHearings(AllocationHearingsToCsoReque
...
             }
         }
+        
+        [HttpPost("HearingCancelled")]
+        [SwaggerOperation(OperationId = "HearingCancelled")]
+        [ProducesResponseType((int)HttpStatusCode.NoContent)]
+        [ProducesResponseType(typeof(string), (int)HttpStatusCode.BadRequest)]
+        public async Task<IActionResult> HearingCancelled(Guid conferenceId)
+        {
+            await _hearingCancelledEventNotifier.PushHearingCancelledEvent(conferenceId);
```
shaed-parkar, line 187: Remove the conference from the cache here?

VideoWeb/VideoWeb/Controllers/InternalEventController.cs
```diff
@@ -171,5 +177,25 @@ public async Task<IActionResult> AllocationHearings(AllocationHearingsToCsoReque
...
+            return NoContent();
+        }
+
+        [HttpPost("HearingDetailsUpdated")]
+        [SwaggerOperation(OperationId = "HearingDetailsUpdated")]
+        [ProducesResponseType((int)HttpStatusCode.NoContent)]
+        [ProducesResponseType(typeof(string), (int)HttpStatusCode.BadRequest)]
+        public async Task<IActionResult> HearingDetailsUpdated(Guid conferenceId)
+        {
+            await _hearingDetailsUpdatedEventNotifier.PushHearingDetailsUpdatedEvent(conferenceId);
```
shaed-parkar, line 197: Good opportunity to force a refresh of the cache here, and then publish the conference object? That way the clients do not have to make an api call to get the object again

shaed-parkar (commented):
Got a few questions with possible suggestions

VideoWeb/VideoWeb.EventHub/Hub/EventHubClient.cs
```diff
@@ -59,12 +61,17 @@ public async Task AddToGroup(string conferenceId)
...
     
-    private async Task AddUserToUserGroup(bool isAdmin)
+    private async Task AddUserToUserGroup(bool isAdmin, bool isStaffMember)
     {
         if (isAdmin)
         {
             await Groups.AddToGroupAsync(Context.ConnectionId, VhOfficersGroupName);
         }
+
+        if (isStaffMember)
```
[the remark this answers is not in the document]
oliver-scott: because we only message VHOs and participants who are part of the hearing. A staff member who is opening the list for monitoring may not be part of the hearing, and so they wouldn't receive the message

VideoWeb/VideoWeb/Controllers/InternalEventController.cs
oliver-scott: I don't think we can unfortunately. The base class for the event handlers (`EventHandlerBase`) fetches the conference from the cache. If we remove the conference, that step will fail

```
public virtual async Task HandleAsync(CallbackEvent callbackEvent)
        {
            SourceConference = await conferenceService.GetConference(callbackEvent.ConferenceId);
            if (SourceConference == null) throw new ConferenceNotFoundException(callbackEvent.ConferenceId);
```

VideoWeb/VideoWeb/Controllers/InternalEventController.cs
oliver-scott: Updated it to force refresh the cache. I agree about publishing the conference object, but not sure we need it yet, as currently the event only needs to act as a trigger to refresh the list on the client, we're not using the contents of the conference object. Maybe later down the line we can hook this up to the redux store and use the conference objects then

VideoWeb/VideoWeb/Controllers/InternalEventController.cs
shaed-parkar: I didn't notice this was extending the EventHandlerBase class. The EventHandler base was designed to handler events from the supplier, not the internal events.
We've exposed internal events to suppliers if they were to send the correct payload to callback.

There is a ticket in the backlog to address VIH-9935, but ideally any internal event handlers shouldn't inherit from the design for the external event handler

shaed-parkar: If you take this strategy to decouple the event handler, you can remove from the cache safely

oliver-scott: good idea. Have pushed 2 commits to address these points:

[First](https://github.com/hmcts/vh-video-web/pull/2219/commits/31eaab9cb67c783e94a02ab91c6b79c87d7f9610) - remove the event handlers for the new events I added, instead move the logic into the notifiers

[Second](https://github.com/hmcts/vh-video-web/pull/2219/commits/e8ba48801e153df2402b1a50b98bddb3fdcd823d) - remove the conference from the cache

VideoWeb/VideoWeb.EventHub/Handlers/NewConferenceAddedEventHandler.cs
```diff
@@ -22,8 +23,10 @@ private async Task PublishNewConferenceAddedMessage(Guid conferenceId)
         {
             foreach (var participant in SourceConference.Participants)
```
shaed-parkar, line 24: you can reduce the nesting if you do. Something Sonar tends to have a whinge about if there's to many layers
SourceConference.Participants.Where(x=>x.Role != StaffMember)

VideoWeb/VideoWeb.EventHub/Handlers/ParticipantsUpdatedEventHandler.cs
```diff
@@ -25,10 +26,14 @@ private async Task PublishParticipantsUpdatedMessage(List<ParticipantResponse> u
     {
         foreach (var participant in participantsToNotify)
         {
-            await HubContext.Clients.Group(participant.UserName.ToLowerInvariant())
-                .ParticipantsUpdatedMessage(SourceConference.Id, updatedParticipants);
-            Logger.LogTrace("{UserName} | Role: {Role}", participant.UserName,
-                participant.Role);
+            // Staff members already receive a message via the staff members group below, so don't message them here as well
+            if (participant.Role != Role.StaffMember)
```
shaed-parkar, line 30:
```suggestion
            foreach (var participant in participantsToNotify.Where(p => p.Role != Role.StaffMember)
```

komaldabhi1509 (approved):

PR #2220, opened 2024-08-27
VIH-10915 connect to signalR AFTER a user is authenticated

shaed-parkar:
### Jira link

VIH-10915

### Change description

* connect to signalR AFTER a user is authenticated

When an unauthenticated user makes a connection to signalR, it seems the same connection id is maintained and the user is not correctly registered to the correct groups. To minimise red herrings in the logs only authenticated users were subscribing to the appropriate groups. To complete this change the ng client app will not initiate the connection until the user is authenticated.


will-craig (approved):
This is a really good spot!

komaldabhi1509 (approved):

PR #2222, opened 2024-08-28
VIH-10586 participant leaves or disconnects

shaed-parkar:
### Jira link

VIH-10586

### Change description

With the new supplier, a non-host participant is now able to leave a hearing.
When a non-host participant clicks the leave button, and confirms:
* Dispatch an action to leave a hearing
  * a side effect will call the BFF:
    * Call the supplier API to transfer to the waiting room
    * Push a message via SignalR with a participant leaving a hearing
* Dispatch a leave hearing success action on success
* Users will subscribe to the new SignalR message
* If the user is a host for the conference, they receive a toast notification

----------
Screenshot 2024-08-28 at 11 15 20

Screenshot 2024-08-28 at 11 11 14


komaldabhi1509 (approved):

marcogagliardi (approved):
LG2M

PR #2236, opened 2024-09-06
VIH-10944 apply yellow border on participant self view when spotlighted

shaed-parkar:
### Jira link

VIH-10944

### Change description

get the logged in participant from redux store
read the spotlight status from store
apply yellow border on self view for any participant that is spotlighted


oliver-scott (approved):

shaed-parkar:
Not needed

PR #2254, opened 2024-10-04
VIH-10689 event hub changes

will-craig:


shaed-parkar (commented):
Because it's a state we need to track, I think you could probably put this into the conference store and then subscribe to a selector.
When you subscribe to the the RecordingState, you could either go down the route of an if recording or else

or

use pipe and filter when it's recording and filter when it's paused

PR #2261, opened 2024-10-08
VIH-10843 fix indentation for master release pipeline

shaed-parkar:
### Jira link

VIH-10843

### DIFF
--- a/azure-pipelines.sds.master-release.yml
+++ b/azure-pipelines.sds.master-release.yml
@@ -21,7 +21,7 @@ parameters:
     buildPushImage: true
     pushHelmChart: true
   - env: 'demo'
-    requireApproval: true
+    requireApproval: false
     buildPushImage: false
     pushHelmChart: false
   - env: 'test'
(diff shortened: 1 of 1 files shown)


azure-pipelines.sds.master-release.yml
```diff
@@ -21,7 +21,7 @@ parameters:
     buildPushImage: true
     pushHelmChart: true
   - env: 'demo'
-    requireApproval: true
+    requireApproval: false
```
oliver-scott, line 24: looks like we might need to change this in admin web too, as it's currently set to true there

at least I take it that is the intention - require approval for all apps in test, but not demo?

oliver-scott (approved):

azure-pipelines.sds.master-release.yml
shaed-parkar: the require approval for demo doesn't actually do anything atm the moment because it's configured to use the same image as dev, so demo updates behind the scenes anyway despite the approval pending. I changed it though it does nothing for no reason other than visually being correct

himesh99 (approved):
lgtm

PR #2264, opened 2024-10-09
VIH-11028 redirect to service error on primary device when participant connects on a secondary device

shaed-parkar:
### Jira link

VIH-11028 

### Change description

There is an outstanding issue involving users who connect to the same hearing with the same account on multiple devices. This typically results in a disconnect of the former connection. However, this overlaps with a retry connection logic which attempts to dial back into the conference after a period of time. This causes both logged in browsers to battle for who is connected and results in a bad experience.

A solution is to capture the reason from the supplier, and if the reason contains the message 'connected on another device' and the participant's call tag, then redirect the user away from the hearing room.

Capture the reason from the callback event and publish it to the participant status message.
### Participant Status Message Enhancements:

* [`VideoWeb/VideoWeb.EventHub/Handlers/Core/EventHandlerBase.cs`](diffhunk://#diff-8ff9406882f340a030bc50a448ea68c7e993d4e98a5598e9445017e1e2e85679L51-R59): Added a `reason` parameter to the `PublishParticipantStatusMessage` method and updated its calls.

### Interface and Unit Test Updates:

* [`VideoWeb/VideoWeb.EventHub/Hub/IEventHubClient.cs`](diffhunk://#diff-ad235ce38d83f07d355a2950d1a446231c7f44bb7f73aac268b6b6d8e013137eL13-R13): Added a `reason` parameter to the `ParticipantStatusMessage` method.

### Dependency Updates:

* [`VideoWeb/VideoWeb/ClientApp/package-lock.json`](diffhunk://#diff-263dc8e35e354652b22954f2bb4387fa80cfed05b9480fdbc284c2bac6c749f7L98-R98): Updated `puppeteer` from version `23.2.0` to `23.5.1` and `@puppeteer/browsers` from version `2.3.1` to `2.4.0`. [[1]](diffhunk://#diff-263dc8e35e354652b22954f2bb4387fa80cfed05b9480fdbc284c2bac6c749f7L98-R98) [[2]](diffhunk://#diff-263dc8e35e354652b22954f2bb4387fa80cfed05b9480fdbc284c2bac6c749f7L5089-R5091)


VideoWeb/VideoWeb.EventHub/Handlers/Core/EventHandlerBase.cs
```diff
@@ -48,15 +48,15 @@ public virtual async Task HandleAsync(CallbackEvent callbackEvent)
         /// <param name="newStatus"></param>
         /// <returns></returns>
```
levibutler, line 49: just passing through; could update summary to include `reason` param

levibutler:
looks good!
did have quick question, why use `callTag` over the `conferenceId` that is already on the event? and used as key for storing the data such as isMuted etc already? Ive probably missed something

VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/store/effects/conference.effects.ts
```diff
@@ -84,10 +87,32 @@ export class ConferenceEffects {
...
+        () =>
+            this.actions$.pipe(
+                ofType(ConferenceActions.updateParticipantStatus),
+                filter(action => action.status === ParticipantStatus.Disconnected),
+                concatLatestFrom(() => [this.store.select(ConferenceSelectors.getLoggedInParticipant)]),
+                switchMap(([action, participant]) => {
+                    if (participant.id !== action.participantId) {
+                        return of();
+                    }
+                    const callTag = participant?.pexipInfo?.callTag ?? this.videoCallService.pexipAPI.call_tag;
```
levibutler, line 100: looks good!
did have quick question, why use callTag over the conferenceId that is already on the event? and used as key for storing the data such as isMuted etc already? Ive probably missed something

shaed-parkar: So I discovered that the call tag isn't set until we get a callback which first arrives when you get into the hearing room. To get the call tag from the waiting room you need to check the video call service

VideoWeb/VideoWeb.EventHub/Handlers/Core/EventHandlerBase.cs
shaed-parkar: done

komaldabhi1509 (approved):

PR #2276, opened 2024-10-21
VIH-11050 disconnect from camera and microphone when user leaves the waiting room

shaed-parkar:
### Jira link

VIH-11050

### HeartbeatService Improvements:
* Added an `initialising` flag to prevent multiple initializations of the heartbeat service. (`VideoWeb/VideoWeb/ClientApp/src/app/services/conference/heartbeat.service.ts`)

### UserMediaStreamService Enhancements:
* Introduced a `closeCurrentStream` method to stop all tracks of the current stream. (`VideoWeb/VideoWeb/ClientApp/src/app/services/user-media-stream.service.ts`)

### VideoCallService Updates:
* Updated the `disconnect` method to call the new `closeCurrentStream` method from `UserMediaStreamService`. (`VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/services/video-call.service.ts`)

### DIFF
--- a/VideoWeb/VideoWeb/ClientApp/src/app/services/user-media-stream.service.ts
+++ b/VideoWeb/VideoWeb/ClientApp/src/app/services/user-media-stream.service.ts
@@ -85,6 +85,12 @@ export class UserMediaStreamService {
         this.activeMicrophoneStreamSubject.next(stream);
     }
 
+    closeCurrentStream() {
+        this.currentStream?.getTracks().forEach(track => {
+            track.stop();
+        });
+    }
+
     private initialiseCurrentStream() {
         this.userMediaService.checkCameraAndMicrophonePresence().then(result => {
             const hasCameraDevices = result.hasACamera;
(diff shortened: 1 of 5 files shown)


VideoWeb/VideoWeb/ClientApp/src/app/services/user-media-stream.service.ts
```diff
@@ -85,6 +85,12 @@ export class UserMediaStreamService {
         this.activeMicrophoneStreamSubject.next(stream);
     }
 
+    closeCurrentStream() {
```
oliver-scott, line 88: could we add a test for this?

oliver-scott (commented):
looks good, just one suggestion

VideoWeb/VideoWeb/ClientApp/src/app/services/user-media-stream.service.ts
shaed-parkar: we can, I've had some real double mocking MediaTracks and it just fails with spys. I;ll have another crack at it but I faced a similar issue in my other or with the v2 media stream service

oliver-scott: don't worry about it if it's going to be a pain, happy to approve for now so we can get it over the line and address later

oliver-scott (approved):

komaldabhi1509 (approved):

PR #2287, opened 2024-10-31
VIH-9653 Updated self test to return OK true/false rather than 200/404

will-craig:
https://tools.hmcts.net/jira/browse/VIH-9653


marcogagliardi (approved):
LG2M

komaldabhi1509 (approved):

PR #2308, opened 2024-11-18
VIH-0000 add job to verify LD key usages

shaed-parkar:
### Jira link

POC

### Change description

Run a job to verify which LD secrets are still in use for safe removal/archiving


shaed-parkar:
replaced by #2321

PR #2318, opened 2024-11-27
VIH-11170 removed vodafone toggle logic from waiting room

will-craig:
### Jira link
https://tools.hmcts.net/jira/browse/VIH-11170

### Change description
Removed use of vodafone toggle from from waiting room, changed conditional code to use the supplier value

### DIFF
--- a/VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/waiting-room-shared/waiting-room-base.component.ts
+++ b/VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/waiting-room-shared/waiting-room-base.component.ts
@@ -18,7 +18,7 @@ import {
     ParticipantStatus,
     Role,
     RoomSummaryResponse,
-    SharedParticipantRoom,
+    SharedParticipantRoom, Supplier,
     VideoEndpointResponse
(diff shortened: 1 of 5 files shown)


VideoWeb/VideoWeb/ClientApp/src/app/waiting-space/waiting-room-shared/waiting-room-base.component.ts
```diff
@@ -820,7 +814,7 @@ export abstract class WaitingRoomBaseDirective implements AfterContentChecked {
         let pexipNode = this.hearing.getConference().pexip_node_uri;
         let conferenceAlias = this.hearing.getConference().participant_uri;
         let displayName = this.participant.tiled_display_name;
-        if (this.needsInterpreterRoom() && !this.vodafoneEnabled) {
+        if (this.needsInterpreterRoom() && this.conference.supplier === Supplier.Kinly) {
```
shaed-parkar, line 817: this section can be removed now because:

1. We won't be using participant VMRs again
2. Vodafone implementation doesn't support it so it might be safer to remove this obsolete code now

PR #2346, opened 2025-01-08
VIH-11132 Hide invite button for screened participants

oliver-scott:
### Jira link
https://tools.hmcts.net/jira/browse/VIH-11132

### Change description
For private consultations, hide the call/invite button when participants are screened
- When the current user is screened from the target participant, or vice versa
- When one of the screening participant pair is already in the consultation

Updates the `participantHasInviteRestrictions` logic, and refactors it into a separate class (ConsultationRules) along with the tests


will-craig (approved):
lg2m

ronaldmalone (approved):

PR #2368, opened 2025-02-05
VIH-11270 Refresh command centre hearing list when hearing is reallocated

oliver-scott:
### Jira link
https://tools.hmcts.net/jira/browse/VIH-11270

### Change description
Fixes an issue where the command centre hearing list doesn't auto refresh when the hearing is reallocated to a different cso than the logged in user.
- Change notifier to publish allocation message to all csos - this way the command centre hearing list will auto-refresh regardless of whether the logged in cso is associated with the hearing
- Only show the notification toast if the allocation is to the currently logged in user
- Refactor frontend tests


VideoWeb/VideoWeb/Helpers/AllocationHearingsEventNotifier.cs
```diff
@@ -21,19 +21,30 @@ public async Task PushAllocationHearingsEvent(UpdatedAllocationJusticeUserDto up
...
-
+    
         var conferences = (await conferenceService.GetConferences(conferenceIds)).ToList();
+        var usernamesToNotify = new List<string>();
         foreach (var conference in conferences)
         {
+            var previouslyAllocatedCsoUsername = conference.AllocatedCso;
+            var newAllocatedCsoUsername = update.AllocatedCsoUsername;
+            
+            if (!string.IsNullOrEmpty(previouslyAllocatedCsoUsername))
```
shaed-parkar, line 32: You could remove the need to check by using the .distinct function on the list

shaed-parkar (approved):
Nice work

VideoWeb/VideoWeb/Helpers/AllocationHearingsEventNotifier.cs
oliver-scott: we still need to continue processing the conference though even if the previous username is null or empty - adding the new cso to the list, updating the conference etc. Or have I misunderstood

shaed-parkar: oh sorry it's on me. what you're doing here is correct. I thought you were checking if the username was already tracked but noticed the loop at the bottom. If CSO A had 2 hearings reassigned from them to CSO B then CSO A needs to updates, 1 per conference/booking re-assigned?

oliver-scott: actually yes you make a good point there. If there are multiple conferences, we need to use a distinct list of usernames to notify in order to avoid pushing the update multiple times to each user.

have pushed a change to handle this. Also updated the test to cover multiple conferences

shaed-parkar (approved):

komaldabhi1509 (approved):

PR #2401, opened 2025-03-26
VIH-11344 self test refresh

shaed-parkar:
### Jira link

VIH-11344

### Change description
The self test component has been replaced with a new version. There were warnings of obsolete functions being invoked on the Pexip API which have been updated.

#### SelfTestV2 component

* New component and updated references to use it
* Better handling of stream and client setup prior to making a call
* Cleaner and better readability in setup **before** different journey types for self test

#### Self Test Score

* Move api calls to retrieve self test score to operate via the Redux Store
* Raise failed test events via effects

#### Video Call Service Refactor

* refactor renegotiate into a queue with rxjs operators to avoid losing a renegotiate request.
* `PexipAPI`'s `stopPresentation` has been deprecated. Switched to using the function `present` with the param `null`


VideoWeb/VideoWeb/ClientApp/src/app/app-routing.module.ts
```diff
@@ -29,6 +29,6 @@ export const routes: Routes = [
 
 @NgModule({
     exports: [RouterModule],
-    imports: [RouterModule.forRoot(routes, { enableTracing: !environment.production })]
+    imports: [RouterModule.forRoot(routes, { enableTracing: environment.production })]
```
oliver-scott, line 32: why the switch to production?

oliver-scott (approved):
looks good, just left a question about the tracing

VideoWeb/VideoWeb/ClientApp/src/app/app-routing.module.ts
shaed-parkar: oh need to revert this! Thanks. it was too noisy during my testing

komaldabhi1509 (approved):

komaldabhi1509 (approved):

PR #2409, opened 2025-04-15
VIH-11343 implement a NoSleepService using the no-sleep-app library

shaed-parkar:
### Jira link

VIH-11343

### Change description

The original implementation was inspired by the same package but does not take advantage of the wake lock api or have fallback options. This library handles this and reduces a dependency on the media stream object which makes unnecessary noise in the logs

### DIFF
--- a/VideoWeb/VideoWeb/ClientApp/src/app/services/no-sleep-v2.service.ts
+++ b/VideoWeb/VideoWeb/ClientApp/src/app/services/no-sleep-v2.service.ts
@@ -14,7 +14,6 @@ export class NoSleepServiceV2 {
     constructor(private logger: Logger) {}
 
     enable() {
-        console.log(`${this.loggerPrefix} no sleep status: ${this.noSleep.isEnabled}`);
         if (!this.noSleep.isEnabled) {
             this.logger.debug(`${this.loggerPrefix} enabling no sleep`);
             this.noSleep.enable();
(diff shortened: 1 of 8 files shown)


VideoWeb/VideoWeb/ClientApp/src/app/services/no-sleep-v2.service.ts
```diff
@@ -0,0 +1,28 @@
...
+})
+export class NoSleepServiceV2 {
+    noSleep = new NoSleepApp();
+
+    private loggerPrefix = '[NoSleepServiceV2] -';
+
+    constructor(private logger: Logger) {}
+
+    enable() {
+        console.log(`${this.loggerPrefix} no sleep status: ${this.noSleep.isEnabled}`);
```
will-craig, line 17: We're injecting the logger, but using console.log here

will-craig, line 17: We're injecting the logger, but using console.log here

shaed-parkar: ha needed to debug in the test, didn't think about that. I swear the lint should have picked it up

will-craig (approved):

oliver-scott (approved):

PR #2417, opened 2025-04-25
VIH-11393 enable mute button only when countdown is complete

shaed-parkar:
### Jira link

VIH-11393

### Change description

Hosts can toggle their local mute before the countdown is complete. Though the Pexip client will honour the request, it does not make a difference since they cannot be heard until the countdown is complete

* Enable mute button only when countdown is complete


marcogagliardi (approved):
LG2M

komaldabhi1509 (approved):

PR #2429, opened 2025-05-19
VIH-11421 enable local mute toggle when countdown is complete or participant is in a private consultation

shaed-parkar:
### Jira link

VIH-11421

### Change description

enable local mute toggle when countdown is complete or participant is in a private consultation


marcogagliardi (approved):
LG2M

PR #2430, opened 2025-05-21
VIH-11357 Add supplier stub

oliver-scott:
### Jira link
https://tools.hmcts.net/jira/browse/VIH-11357


shaed-parkar (commented):
did you check where the heartbeats are going?

shaed-parkar (commented):
It may also be worth considering to change the video call service to exit early if the supplier is a stub as to avoid the call being set up and the participant can sit in the WR without crashing

shaed-parkar (approved):

komaldabhi1509 (approved):

